PR #386, opened 2018-12-05
CSS bug + Refactoring post on/off

mridulnagpal:


mridulnagpal:
@natejenkins @spMohanty This PR resolves the CSS bug as well as combines 2 files `app/views/leaderboards/_post_challenge_off.html.erb` and `app/views/leaderboards/_post_challenge_on.html.erb` to 1 file `app/views/leaderboards/_post_challenge.html.erb`

app/assets/stylesheets/_forms.scss
```diff
@@ -294,7 +294,6 @@ input.switch:checked ~ label:after {
 
 form.leaderboard {
   margin-top: 20px;
-  width: 200px;
```
natejenkins: you mentioned that this was added for another part of the codebase.  Where was that exactly?

natejenkins: can we add an extra class so that the existing behavior is maintained for submissions or is that css also broken?

PR #736, opened 2018-12-07
simple fix for baseline leaderboard fix in view layer, does not alter DB

### LINKED ISSUE #1084: Challenge rounds view need to be improved
We should have design modification for the rounds edit page, which isn't rendered well right now.

For this we will need Jason inputs in future, but for now we can have slight modifications from our side:

1. Clear seperator between multiple rounds, so it is visually better
2. Collapsable rounds will be good to have, so the settings are contained/hidden which organiser don't want to see
3. Rounds are not being removed i.e. remove button only removed the last added round and second last (and so on) can't be removed.

natejenkins:
fixes #1084

### DIFF
--- a/app/concepts/leaderboard/cells/table_row.rb
+++ b/app/concepts/leaderboard/cells/table_row.rb
@@ -12,6 +12,10 @@ class Leaderboard::Cell::TableRow < Leaderboard::Cell
     model
   end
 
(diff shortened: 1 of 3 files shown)


mridulnagpal:
Pulled the branch and checked, seems to work fine! Should we merge?

PR #751, opened 2018-12-15
reorder challenges via drag-and-drop (work in progress)

natejenkins:
Overall this is pretty good @mridulnagpal .  We need to add a link to it somewhere and my suggestion is under the admin interface which uses ActiveAdmin.  Why don't you see if you can add a batch action for challenges (https://activeadmin.info/9-batch-actions.html)

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -35,6 +35,25 @@ class ChallengesController < ApplicationController
     end
   end
 
+  def reorder
+    authorize Challenge
+    @challenges = policy_scope(Challenge)
+  end
+
+  def assign_order
+    authorize Challenge
+    @final_order = params[:order].split(",")
+
+    len = @final_order.length
+
+    @final_order.each do |i|
+      Challenge.friendly.find(i).update(featured_sequence: len)
+      len = len - 1
+    end
+
+    redirect_to challenges_url
+  end
+
   def show
     if !params[:version]  # dont' record page views on history pages
       @challenge.record_page_view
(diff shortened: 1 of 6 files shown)


app/views/challenges/assign_order.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class='content'>
+	<div class="masthead">
+		<div class="row">
+			<h2>Reorder Challenges</h2>
+		</div>
+	</div>
+	<div class="row">
+		<div class="section-header">
+			<h6 style="color: green;">Success</h6>
```
natejenkins: Instead of rendering a Success  indicator you should redirect to the challenges route so we can immediately see the result of the reordering

app/controllers/challenges_controller.rb
```diff
@@ -36,7 +36,23 @@ def index
...
+    @final_order = params[:order].split(",")
+
+    len = @final_order.length
+
+    @final_order.each do |i|
+      Challenge.friendly.find(i).update(featured_sequence: len)
+      len = len - 1
+    end
+    
+    render 'challenges/assign_order'
```
natejenkins: this line is unnecessary, it is implied in rails

app/views/challenges/assign_order.html.erb
mridulnagpal: Done

mridulnagpal:
@natejenkins So I read about batch actions, apparently I couldn't find a batch action which simply renders a controller, i.e challenges#reorder. I am looking for to include it in active admin somehow. Let me know if you have any ideas how to go forward.

mridulnagpal:
@natejenkins Got it, concerning the tabs I'll tweak my editor settings. Why are the tests failing? They seem to pass in my local

PR #753, opened 2018-12-18
Trueskill (Work in progress)

mridulnagpal:
Rating model added


Gemfile.lock
```diff
@@ -948,7 +955,7 @@ DEPENDENCIES
   webpacker (~> 2.0)
 
 RUBY VERSION
-   ruby 2.5.1p57
+   ruby 2.3.7p456
```
natejenkins: definitely not going to downgrade ruby, sorry

mridulnagpal: Sudo access must have done this

Issue #764, opened 2019-01-05
Rename HOST env variable to DOMAIN


spMohanty:
The Env variable `HOST` is a bit misleading. Its primarily used to create self-referential links to include in emails, etc. Its better called as`DOMAIN`

PR #765, opened 2019-01-05
Addresses #764 - Renames Env Var HOST to DOMAIN_NAME

spMohanty:


spMohanty:
@natejenkins : Fair enough. But given the context in which it is used, it could also technically reside at a subdomain. Say : `https://<some_domain.com>/aicrowd`
But lets go with DOMAIN_NAME now to avoid confusion, and revisit this later.

PR #818, opened 2019-02-06
Add Participation Terms workflow

nemaniarjun:


app/concepts/participation_terms/views/show.erb
```diff
@@ -0,0 +1,32 @@
...
+    <div class="row">
+
+      <!-- sub-nav – vertical -->
+      <div class="col-lg-2 sub-nav-vertical-container">
+      </div>
+      <!-- /sub-nav – vertical -->
+
+      <!-- md content -->
+      <div class="col-md-8 col-lg-7">
+		<%= challenge.toc_acceptance_text %></br>
```
natejenkins: we probably shouldn't repeat this four times

app/controllers/dataset_files_controller.rb
```diff
@@ -72,6 +67,19 @@ def set_challenge
     @challenge = Challenge.friendly.find(params[:challenge_id])
   end
 
+  def check_participation_terms
+    if current_participant && (current_participant.admin? || @challenge.organizer_id == current_participant.organizer_id)
```
natejenkins: I added this check for admins and organizers because it seems reasonable that they should not have to be challenge_participants and accept the terms

app/controllers/submissions_controller.rb
```diff
@@ -129,6 +133,19 @@ def set_challenge
       @challenge = Challenge.friendly.find(params[:challenge_id])
     end
 
+    ### TODO: extract this out into a shared function with the
+    ### dataset_files_controller and also determine if it needs to
+    ### be bypassed for organizers and admins.
+    def check_participation_terms
```
natejenkins: I don't like the function duplication here.  It probably should be identical to that found in the dataset_files_controller.  It can be extracted out into a shared function.

app/services/grader_service.rb
```diff
@@ -8,6 +8,19 @@ class GraderService
 
   def initialize(submission_id:)
     @submission = Submission.find(submission_id)
+    challenge = @submission.challenge
+    participant = @submission.participant
+    challenge_participant = challenge
+      .challenge_participants
+      .find_by(participant_id: participant.id)
+
+    if challenge_participant.blank? or !challenge_participant.accepted_dataset_toc
```
natejenkins: this isn't code I know very well but again the question of admins and organizers comes up, should there be a check for them to bypass this requirement here?

app/views/challenges/show/_subnav.html.erb
```diff
@@ -79,6 +79,9 @@
             new_challenge_submission_path(challenge),
             class: "btn btn-secondary pull-right" %>
     <% end %>
+    <%# if !challenge_participant or !challenge_participant.accepted_dataset_toc %>
```
natejenkins: I commented this out because the parts of the codebase which call this partial are not related to submissions or to the dataset files.  I also compared this to the behavior in the redesign branch.  Do you agree @nemaniarjun ?

app/services/grader_service.rb
nemaniarjun: Nah, we don't expect the admins and organizers to make submissions so that does not seem needed.

app/views/challenges/show/_subnav.html.erb
nemaniarjun: I agree. I think for the redesign branch are aim should be to remove all partials or all concept as they seem redundant concepts.
concepts cells > partials for me

app/controllers/submissions_controller.rb
nemaniarjun: I thought of the same thing, but let it be for now because I am not that experienced in rails and we are going to overwrite this for redesign branch anyway,

app/controllers/dataset_files_controller.rb
nemaniarjun: seems great

app/concepts/participation_terms/views/show.erb
nemaniarjun: yeah that was just for the seed

db/migrate/20181030213518_toc.rb
```diff
@@ -1,6 +1,6 @@
 class Toc < ActiveRecord::Migration[5.2]
   def change
-	  add_column :challenges, :toc_acceptance_text, :string 
+	  add_column :challenges, :toc_acceptance_markdown, :text
```
natejenkins: I didn't realize at first that this was already a migration in the master branch.  In general you should not modify old migrations, you need to think of them as immutable once committed to master since Rails will not run migrations it has already run before.  You need to write a new migration which modifies a column.

app/controllers/dataset_files_controller.rb
```diff
@@ -3,21 +3,16 @@ class DatasetFilesController < ApplicationController
   before_action :set_dataset_file,
     only: [:destroy, :edit, :update]
   before_action :set_challenge
+  before_action :check_participation_terms, only: [:index, :show]
+  # Before we allow user to view files we will check if they have agreed to the toc
```
natejenkins: also I find it a little confusing having comments after a line of code rather than before a line of code.

PR #867, opened 2019-03-05
Show media on leaderboard according to challenge settings

nemaniarjun:


nemaniarjun:
@spMohanty, I am not sure if this was a valid commit! Maybe we might have to roll back, this PR was pretty stale and untested.

spMohanty:
@nemaniarjun : Its not deployed yet. I am working on it now. Will test it and then deploy. 
If you are around to give a hand, that would be great too !

nemaniarjun:
Unfortunately traveling right now, so don't have access to the internet.
Would be able to help in approx 12 hours.

PR #878, opened 2019-03-07
Update .travis.yml

natejenkins:
give some more time for localstack to spin up to see if it fixes the travis builds


spMohanty:
@natejenkins : This test is not relevant anymore. Organizers dont use the API anymore. So the test can be removed.

Issue #958, opened 2019-06-01
Participants count for competition and other statistics field update slowly


skbly7:
We are getting lot of values like participants count for a competition populated as job here:
https://github.com/AIcrowd/AIcrowd/blob/1dd930695f0aa0aa9602633e4a3fe2bebefa9d10/app/jobs/update_challenge_stats_job.rb#L22

This job seems to be running infrequently which is leading to lower counts on website. For example, we found participants count on website to be 310 instead of 422 for a competition and so on.

seanfcarroll:
It does [_seem_](https://github.com/AIcrowd/AIcrowd/blob/bd90bfbea008e9e70fec6ce9269ec2b247ee1f72/app/controllers/challenges_controller.rb#L9) to be running on every page load of the challenge ... possibly a little too often but certainly frequent.

If the stats are not properly updating I'd suggest reviewing the code of the job itself. This [SQL](https://github.com/AIcrowd/AIcrowd/blob/1dd930695f0aa0aa9602633e4a3fe2bebefa9d10/app/jobs/update_challenge_stats_job.rb#L23-L32) looks suspicious.

nemaniarjun:
@spMohanty  @marcelsalathe 
Do we want the count to be of all the participants who have accepted "any" version of the rules? or just the latest version?

marcelsalathe:
Any

nemaniarjun:
@spMohanty @bkDJ 
The sql in that update job seems to be wrong. 
If we want to display the number of people who have accepted any version of the challenge we can just use challenge.challenge_participants (the same one we use on the participants page)

PR #966, opened 2019-06-05
Fix count for participants, and show post-challenges switch for completed challs only

nemaniarjun:
Removed the SQL job and using challenge.challenge_participants
for the count. Also showing all the challenge_participants
on the participants tab instead of the ones that accepted
only the latest terms.

Fixes https://github.com/AIcrowd/AIcrowd/issues/958


nemaniarjun:
@spMohanty should I merge this?

PR #977, opened 2019-06-29
Leaderboard Note and 500 on /participants

nemaniarjun:
@spMohanty please check my comments in both the issues and confirm if the way I am handling is the desired way.

Main discussion points: Do we wrap the note in class=update or have the challenge organizers add it to the markdown manually?

When a participant tries to access /participants do we redirect to the home page or somewhere else?

### DIFF
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -25,11 +25,17 @@ Paloma.controller('Challenges', {
     })
   },
   show: function(){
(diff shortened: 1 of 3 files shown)


nemaniarjun:
@spMohanty is this ready to be merged?

nemaniarjun:
@spMohanty I am merging these into master.

The discussion may be hidden due to the force push, you can read it here
https://github.com/AIcrowd/AIcrowd/commit/08502523362bad2bde7471aed4e3a950b2cb5f63#comments

PR #984, opened 2019-07-21
Jobs page, leaderboard note, mobile stats, leaderboard post-challenge filter

nemaniarjun:


nemaniarjun:
Fixing Tests in 5 mins!

jsnrynlds:
Do we need 'AIcrowd' under the job title? Seems a bit unnecessary.

I would change 'Hyderabad, India :: Remote OK' to: 'Hyderabad, India or remote'

nemaniarjun:
As I mentioned if we add the organization name to the design we can use it
for posting jobs for other organizations as well.

Also we would need to add contact info fields to the design.

jsnrynlds:
> As I mentioned if we add the organization name to the design we can use it
for posting jobs for other organizations as well.

Okay, makes sense but the design will need tweaking.

> Also we would need to add contact info fields to the design.

On the [detail page](https://aicrowd-redesign.herokuapp.com/job_postings/aicrowd-rails-dev)?

nemaniarjun:
Yes, the contant info on the detail page. 
If we can have svgs for remote and full time / part time, I will add them to the design. but that may cause the boxes on the index page to be of different size.

nemaniarjun:
@jsnrynlds Ping! I have fixed the remote text and changed the slug to just 'jobs'

nemaniarjun:
Also @marcelsalathe @jsnrynlds 
The text on this page: https://aicrowd-redesign.herokuapp.com/jobs
is hardcoded. Can we have the actual content we should display there? 

Thanks!

marcelsalathe:
Interested in working at AIcrowd? Our open positions are below. AIcrowd is a "mostly remote" company - our HQ is in the Geneva area in Switzerland, but most of us work from wherever we happen to live, and the way we work strongly reflects that.

jsnrynlds:
Wondering if we're going to be posting jobs from other organisations here, too? If so we should probably add a sentence referring to this or tweak the intro text.

jsnrynlds:
> Also we would need to add contact info fields to the design.

Can we also clarify whether we need to include contact details on the [open position detail page](https://aicrowd-redesign.herokuapp.com/jobs/aicrowd-rails-dev)? (Not sure it's necessary with the 'Apply now' button.)

nemaniarjun:
@marcelsalathe can you answer the 2 questions?

1. Do we support other organizations? 
2. Do we display contact information. 

For me, both are nice to have, but we don't need them now.

spMohanty:
@nemaniarjun : I think we should have the ability to support job ads from other organisations. But when rendering the job positions, the availabilities from AIcrowd will ofcourse take a precedence (render a block of all AIcrowd job positions first) ? @jsnrynlds  @marcelsalathe ?

And I dont think we should display contact information. The reasoning being, people who want to be contacted, can include their information in the job ad text. But if we automatically start including the contact information of the users who post (eventually), then some might not realize that their contact details are public without their active intent.

nemaniarjun:
Okay!
I will update accordingly.

@Jason that just leaves the need of a svg to show beside the name of the
organization and if/how we should display the remote / full time tags (on
both index and detail)

marcelsalathe:
I would drop the idea of other organizations for now. We can in the future think about it, but for now this is our own hiring page.

No contact information - juts the ability to link to an external recruitment platform

nemaniarjun:
Dropped the Idea for other organizations, pushed an update to never show on any pages, but it is present in the model for now, so we can use it if needed later. 
Jobs part is more or less complete now. @marcelsalathe please check it on https://aicrowd-redesign.herokuapp.com/jobs

@spMohanty Also pushed a fix for the leaderboard note to the same PR you can see it at https://aicrowd-redesign.herokuapp.com/challenges/unity-obstacle-tower-challenge/leaderboards

@bkDJ Fixed a push for the missing leaderboard stats on mobile.


@jsnrynlds can you confirm all 3 if any changes are required or not?

jsnrynlds:
Re jobs, I think we're better off just including 'Full-time' or 'Part-time' in the job title (as it is, although we should hyphenate these).

Other changes look fine to me @nemaniarjun – thanks.

PR #998, opened 2019-08-25
Migration model for keeping logs for which crowdai participant is mapped to which aicrowd participant 

mridulnagpal:
crowdai_participant_id -> aicrowd_participant_id mapping

### DIFF
--- a/app/models/migration_log.rb
+++ b/app/models/migration_log.rb
@@ -1,0 +1,2 @@
+class MigrationLog < ApplicationRecord
+end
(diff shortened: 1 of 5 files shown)


mridulnagpal:
@bkDJ The migration_mapping table and crowdai_participant_mappings tables will be different or one table only. I am currently keeping mappings of submissions, etc. to participants in a separate table and logging the participants who have migrated in a separate table. This PR is for the migration table specified in the spec doc. I was thinking of filling pairs of crowdai to aicrowd pid mappings in this table after a user has migrated to keep a log. What do you think about this?

nemaniarjun:
Closing in favor of a new PR for all migration related things!

PR #1007, opened 2019-09-16
Challenge: Add Team Freeze Time

nemaniarjun:


app/controllers/challenges_controller.rb
```diff
@@ -164,6 +164,7 @@ def challenge_params
             :teams_allowed,
             :hidden_challenge,
             :max_team_participants,
+            :freeze_time,
```
spMohanty: @nemaniarjun : Lets use `team_freeze_time` consistently to avoid confusion in the future.

PR #1035, opened 2019-11-14
Leaderboard Calculation: Order Refactor \& Submission views formatting

nemaniarjun:
This commit refactors the Leaderboard Calculation Service to properly
handle the orders that are defined into the challenge.
Also refactors the submission detail and submission list views
to show the correct formatted value.

### DIFF
--- a/app/concepts/leaderboard/cells/table_row.rb
+++ b/app/concepts/leaderboard/cells/table_row.rb
@@ -37,7 +37,7 @@ class Leaderboard::Cell::TableRow < Leaderboard::Cell
   end
 
   def participants
-    @participants ||= begin
+    temp_participants = begin
       if entry.try(:submitter_type) == 'Team'
         entry&.team&.participants&.to_a
       else
@@ -45,7 +45,8 @@ class Leaderboard::Cell::TableRow < Leaderboard::Cell
       end
     end
 
-    @participants = [] if @participants.nil? || @participants.include?(nil)
+    temp_participants = [] if temp_participants.nil? || temp_participants.include?(nil)
+    return temp_participants
   end
 
   def top_rows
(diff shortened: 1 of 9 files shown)


app/services/calculate_leaderboard_service.rb
```diff
@@ -39,64 +36,16 @@ def call
...
+                      .order(created_at: :desc)
+                      .limit(1)
+                      .first
     window_border = most_recent.created_at - @round.ranking_window.hours
     return "'#{window_border.to_s(:db)}'"
   end
 
-  def get_order_by
-    challenge = @round.challenge
-    if (challenge.secondary_sort_order_cd.blank? || challenge.secondary_sort_order_cd == 'not_used')
```
skbly7: What if blank need to be added back.

app/concepts/leaderboard/cells/table_row.rb
```diff
@@ -12,13 +12,8 @@ def entry
     model
   end
 
-  def formatted_score
```
skbly7: `formatted_score` is used in baseline_row.erb

app/services/calculate_leaderboard_service.rb
```diff
@@ -395,4 +355,33 @@ def set_leaderboard_sequences(leaderboard_type:)
...
+  end
+
+  def order_by(use_display: false)
+    challenge = @round.challenge
+
+    if challenge.latest_submission == true
+      return 'updated_at desc'
+    end
+
+    score_sort_order ||= sort_map(challenge.primary_sort_order_cd)
```
skbly7: What will happen when “nil” is returned back (because of Not Used as value or so)?

nemaniarjun: It can never be nil for primary score. New challenges have that as default.

PR #1037, opened 2019-11-14
Leaderboard View: Fix bug in leaderboard view causing 500s

nemaniarjun:
https://rollbar.com/aicrowd-production/aicrowd-production/items/201

### DIFF
--- a/app/concepts/leaderboard/cell.rb
+++ b/app/concepts/leaderboard/cell.rb
@@ -78,7 +78,7 @@ class Leaderboard::Cell < Template::Cell
   end
 
   def formatted_value(value)
-    format("%.#{entry.challenge_round.score_precision}f", value || 0)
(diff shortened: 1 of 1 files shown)


app/concepts/leaderboard/cell.rb
```diff
@@ -78,7 +78,7 @@ def organizer_badge?(challenge_id:,participant_id:)
   end
 
   def formatted_value(value)
-    format("%.#{entry.challenge_round.score_precision}f", value || 0)
+    format("%.#{challenge_round.score_precision}f", value || 0)
```
spMohanty: If the `value` is non numeric, then this will also cause an error. So we should check if the value is numeric and then do this formatting, else we should just render as is.

nemaniarjun:
It's always going to be numeric as it is called with score and score
secondary.

PR #1039, opened 2019-11-15
Fix Team URL aliasing issues

bkDJ:
Team URLs now include challenge name
* moved around files so they make sense with new paths
* old URLs will redirect to first team with same name
* updated all links (e.g. emails) and redirects (e.g. create/edit actions or permissions failures)
* updated tests
* updated ActiveAdmin (can see teams of specific challenges, or all teams, fixed URLs, etc)


app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,15 @@
+# This is a legacy controller.
+# A team without a specified challenge might have name aliasing issues
+# so we redirect to the correct page
+class TeamsController < ApplicationController
```
skbly7: Does this be here?

```
before_action :authenticate_participant!, except: [:show]
```

nemaniarjun:
Build is failing

app/controllers/teams_controller.rb
bkDJ: The only action in this controller is `show`. So a `before_action` callback that would only happen for non-`show` actions is unnecessary.

nemaniarjun: @skbly7 also the authenticate_participant callback is used to make sure that a user is logged in, we dont need that here.

PR #1057, opened 2020-01-10
Use Gitlab Rubocop code styles and apply them to codebase

piotrekpasciak:
Changes generated by `rubocop --safe-auto-correct` command


nemaniarjun:
@piotrekpasciak I would like to have a rule that makes it a requirement to have the return statement. The absence just doesn't sit right with me while reading the code.

piotrekpasciak:
@nemaniarjun Unfortunately, I wasn't able to find linter rule that would enforce return statement:
https://github.com/rubocop-hq/rubocop/blob/master/manual/cops.md 

I was only able to disable rule that was removing return statement in last line of code, but there is no auto-correction for that:
https://github.com/AIcrowd/AIcrowd/pull/1057/files#diff-11a0d7906801d9dea0eccb85667ad811R71

The standard way in Ruby styleguide and Ruby community is to remove redundant `return` keywords and for this reason Rubocop seems to be missing this rule. We'll have to add it manually.

app/admin/dashboard.rb
```diff
@@ -1,9 +1,9 @@
 ActiveAdmin.register_page "Dashboard" do
+  menu priority: 1, label: proc { I18n.t("active_admin.dashboard") }
```
nemaniarjun: These seems like an unintended change.

app/concepts/dataset/cells/list_detail.rb
```diff
@@ -25,61 +24,56 @@ def expiring_url
...
     end
   end
 
   def s3_file_obj
     s3_key = dataset_file.dataset_file_s3_key
     return nil if s3_key.nil?
+
     s3_file_obj = Aws::S3::Object.new(bucket_name: ENV['AWS_S3_BUCKET'], key: s3_key)
-    if s3_file_obj && s3_file_obj.key && !s3_file_obj.key.blank?
+    if s3_file_obj&.key && !s3_file_obj.key.blank?
```
nemaniarjun: ```if s3_file_obj&.key && !s3_file_obj.key.blank?```
this can be:  ```unless s3_file_obj&.key&.present?```

app/concepts/dataset/cells/list_detail.rb
```diff
@@ -25,61 +24,56 @@ def expiring_url
   end
 
   def file_title
-    if dataset_file.title.present?
-      dataset_file.title
-    else
-      dataset_file.description
-    end
+    dataset_file.title.presence || dataset_file.description
```
nemaniarjun: if title is present will the return `true` or the `dataset_file.title`?

app/concepts/task_dataset/cells/list_detail.rb
```diff
@@ -22,47 +21,42 @@ def expiring_url
...
     s3_key.split('/')[-1]
   end
 
   def file_title
-    if task_dataset_file.title.present?
-      task_dataset_file.title
-    else
-      task_dataset_file.description
-    end
+    task_dataset_file.title.presence || task_dataset_file.description
```
nemaniarjun: Same as above, we really should remove trailblazer completely, the code is not even DRY for the cells.

app/concepts/task_dataset/cells/list_detail.rb
```diff
@@ -22,47 +21,42 @@ def expiring_url
...
+    ext&.upcase
   end
 
   def s3_file_obj
     s3_key = task_dataset_file.dataset_file_s3_key
     return nil if s3_key.nil?
+
     s3_file_obj = Aws::S3::Object.new(bucket_name: ENV['AWS_S3_BUCKET'], key: s3_key)
-    if s3_file_obj && s3_file_obj.key && !s3_file_obj.key.blank?
+    if s3_file_obj&.key && !s3_file_obj.key.blank?
```
nemaniarjun: ```unless s3_file_obj&.key&.present?```

app/controllers/articles_controller.rb
```diff
@@ -27,31 +26,26 @@ def show
...
 
-
   def new
     @article = Article.new
     authorize @article
   end
 
-  def edit
-  end
+  def edit; end
```
nemaniarjun: I am unsure about this change. Do you prefer this @piotrekpasciak

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,45 +1,37 @@
...
-      redirect_to(root_path, flash: {error: 'Error while processing, contact help@aicrowd.com'}) and return
-    end
+    redirect_to(root_path, flash: { error: 'Error while processing, contact help@aicrowd.com' }) && return if @cid.nil?
 
     @migrate_service = MigrateUserService.new(crowdai_id: @cid, aicrowd_id: current_participant.id)
 
-    if @migrate_service.check_migrated
-      redirect_to(root_path, flash: {error: 'Participant Already Migrated'}) and return
-    end
+    redirect_to(root_path, flash: { error: 'Participant Already Migrated' }) && return if @migrate_service.check_migrated
```
nemaniarjun: Unsure about this change. is ```redirect && return``` the same as ```redirect and return``` also wouldn't we prefer using `and` over `&&`

app/policies/application_policy.rb
```diff
@@ -36,14 +36,14 @@ def destroy?
...
-    issues = RequestStore.store[cache_key]
+    issues    = RequestStore.store[cache_key]
     if issues.nil?
       issues = {}
       # `block.call` returns a conditions hash,
       #   which we reduce into an array of keys for which the value is truthy,
       #   of which the first is considered the most important
-      issues[:list] = block.call.reduce([]) { |acc,(k,v)| acc << k if v; acc }
-      issues[:sym] = issues[:list].first
+      issues[:list]                 = yield.each_with_object([]) { |(k, v), acc| acc << k if v; }
```
nemaniarjun: Needs manual checking.

app/services/s3_service.rb
```diff
@@ -16,15 +15,15 @@ def valid_key?
...
+    if s3_file_obj&.key && !s3_file_obj.key.blank?
       return s3_file_obj.presigned_url(:get, expires_in: 7.days.to_i)
     else
       return nil
     end
   end
 
   def public_url
-    if s3_file_obj && s3_file_obj.key && !s3_file_obj.key.blank?
+    if s3_file_obj&.key && !s3_file_obj.key.blank?
```
nemaniarjun: probably can be refactored better. We can leave these kind of issues for now I think, as they should still work,

app/concepts/dataset/cells/list_detail.rb
piotrekpasciak: Actually it's going to be `if  s3_file_obj&.key.present?`.

`present?` method can be called on object belonging to `Nil` class.

app/concepts/dataset/cells/list_detail.rb
piotrekpasciak: Yes it will, because `presence` method will return value of `dataset_file.title` only if it's not `nil` nor empty string.

app/concepts/task_dataset/cells/list_detail.rb
piotrekpasciak: I support the idea of removal of Trailblazer cells, but let's do that in future in separate PR.

app/concepts/task_dataset/cells/list_detail.rb
piotrekpasciak: `if s3_file_obj&.key.present?`

app/controllers/articles_controller.rb
piotrekpasciak: Yes, I prefer this one line style of empty method. That's a standard approach based on Ruby style guide.

app/controllers/crowdai_migration_controller.rb
piotrekpasciak: In this context, there will be no difference how code will work, but there is very fragile difference between `and` and `&&` operators:
https://www.compassitesinc.com/difference-between-and-and-and-and-or-in-ruby/

We definitely prefer to use `&&` operator as reliable and predictable. Precedence of `and` is usually confusing and can lead to bugs that are difficult to track.

app/policies/application_policy.rb
piotrekpasciak: Screenshot 2020-01-14 at 10 08 35

config/initializers/client_side_validations.rb
```diff
@@ -12,7 +12,7 @@
 # <label for="#{instance.send(:tag_id)}" class="message"></label>
 #
 ActionView::Base.field_error_proc = proc do |html_tag, instance|
-  if html_tag =~ /^<label/
```
skbly7: Is `if /^<label/.match?(html_tag)` valid or it need to be `if html_tag.match?(/^<label/)`?

spec/cells/challenge/cell_spec.rb
```diff
@@ -2,7 +2,9 @@
   describe 'cell can be instantiated' do
     let!(:challenge) { create :challenge }
     let!(:participant) { create :participant }
-    subject { cell(described_class, challenge, current_participant: participant ) }
-    it { expect( subject).to be_a Challenge::Cell }
+
+    subject { cell(described_class, challenge, current_participant: participant) }
+
+    it { expect(subject).to be_a described_class }
```
skbly7: It need to be `Challenge::Cell` or `described_class` here?

skbly7: Same for other cell files, in which code has been changed to `described_class`

nemaniarjun: described class refers to the class being tested in this case 'Challenge:Cell'

piotrekpasciak: Standard approach is to use `desribed_class` as it's easier to modify/refactor the code later.

config/initializers/client_side_validations.rb
piotrekpasciak: I'll  test it.

piotrekpasciak: Screenshot 2020-01-14 at 14 21 19

I don't see any difference between both versions.

PR #1060, opened 2020-01-13
Remove commented code

piotrekpasciak:


app/assets/javascripts/application.js
```diff
@@ -36,7 +36,6 @@
 //= require modules/direct_s3_upload
 //= require modules/mentions
 //= require modules/modals
-//= require modules/tooltips
```
skbly7: Is this not being used?

skbly7: I see tooltip.js has been removed, but there are some places having tooltip reference:
https://github.com/AIcrowd/AIcrowd/search?q=tooltip&unscoped_q=tooltip

Just making sure those are being handled via different require/js file.

app/assets/stylesheets/_tables.scss
```diff
@@ -19,9 +19,6 @@ table {
 		color: $color-grey-54;
 		a {
 			color: $color-grey-54;
-			&:hover {
```
skbly7: This might exist to override hover styling in some parent class?

app/assets/javascripts/application.js
nemaniarjun: This changes nothing @skbly7 the imported file was having no effect. nothing to worry about here.

app/assets/stylesheets/_tables.scss
piotrekpasciak: I  thought that options need to be defined to override anything. I'll put it right back to make sure I didn't modify anything.

PR #1062, opened 2020-01-13
Tidy up tests in spec/controllers folder

piotrekpasciak:


spec/controllers/clef_tasks_controller_spec.rb
```diff
@@ -71,51 +61,26 @@
...
 
-        it "redirects to the clef_task" do
-          #  put :update, params: { organizer_id: organizer.id, id: task_1.to_param, clef_task: valid_attributes}
-          #  expect(response).to redirect_to(clef_task)
-        end
-      end
-
-      context "with invalid params" do
-        it "assigns the clef_task as @clef_task" do
-          # put :update, params: { organizer_id: organizer.id, id: task_1.to_param, clef_task: invalid_attributes}
```
nemaniarjun: instead of removing these tests altogether, we should rewrite them.

spec/factories/participant_factory.rb
```diff
@@ -67,16 +63,22 @@
     end
 
     trait :clef_incomplete do
-      address { nil }
+      address nil
     end
 
     trait :clef_complete do
-      address { 'test test test' }
+      address     { 'test test test' }
```
nemaniarjun: We should be using FFaker for these values

spec/controllers/clef_tasks_controller_spec.rb
piotrekpasciak: I've rewritten specs from spec/controllers/clef_tasks_controller_spec.rb.

spec/factories/participant_factory.rb
piotrekpasciak: I've updated factory to use FFaker here.

piotrekpasciak:
@nemaniarjun I've rewritten every commented test except:
- spec/controllers/crowdai_migration_controller_spec.rb -  this is empty file
- spec/controllers/admin/* - files with one line placeholder that doesn't check anything

Is that sufficient for this PR?

PR #1075, opened 2020-01-15
Revert "Hound: Add Rubocop"

nemaniarjun:
This reverts commit 4912d50e00ef9b499becd8e7b23dcc271afac9b7.


nemaniarjun:
Basically to sum up all I tried:
- Different GitHub Actions that are already available. 
- Hound CI 
- CodeClimate
- travis-ci 

Ran into issues with like no support for client configured plugins that need to be installed, i.e. `gitlab-styles`or other tough to configure issues. 
Closest I came was using https://github.com/Freshly/Octocop , but even that did not work due to some dependency issues.

Right now with this PR, travis is going to run rubocop and the build will fail if rubocop encounters an "error" level correction. 

@skbly7 @piotrekpasciak

piotrekpasciak:
@nemaniarjun

PR #1076, opened 2020-01-16
Fix permission issue for CLEF tasks when on organizer page

ieggel:
@nemaniarjun 
Discovered a bug where ordinary participants were able to see the dataset files on the organizer page, even whey they did not sign an EUA. This has been fixed now.
This PR can be safely merged, only changed permissions in some clef_specific views.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -647,44 +647,6 @@ ActiveRecord::Schema.define(version: 2020_01_15_123518) do
     t.integer "participant_id"
   end
 
-  create_table "migration_mappings", force: :cascade do |t|
-    t.string "source_type"
-    t.integer "source_id"
-    t.integer "crowdai_participant_id"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
-
-  create_table "merit_activity_logs", force: :cascade do |t|
-    t.integer "action_id"
-    t.string "related_change_type"
-    t.integer "related_change_id"
-    t.string "description"
-    t.datetime "created_at"
-  end
-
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -655,6 +655,36 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "merit_activity_logs", force: :cascade do |t|
```
piotrekpasciak: This should not be part of this PR. These tables were removed from the repository lately. Could you make sure you've run following migration:
https://github.com/AIcrowd/AIcrowd/blob/master/db/migrate/20200115123518_remove_merit_database_tables.rb

If that doesn't help, can you rebuild your database in development environment to get rid . of these changes?

piotrekpasciak: These tables are still in schema.rb on your branch. Could you remove them?

ieggel: I removed them now. Hope everything is OK now.

db/schema.rb
```diff
@@ -647,14 +647,6 @@
     t.integer "participant_id"
   end
 
-  create_table "migration_mappings", force: :cascade do |t|
```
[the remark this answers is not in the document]
ieggel: My bad. Really sorry for this. Will push in a few seconds.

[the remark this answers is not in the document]
piotrekpasciak: No need to say sorry. We do Code Review to catch issues like that :)

piotrekpasciak:
@ieggel I've approved it. Feel free to merge the PR whenever you want to.

Issue #1002, opened 2019-09-06
Code Burn 🔥 


nemaniarjun:
I think I have enough knowledge about the codebase to delete all the code that we are not using, figure out places we dont have tests for, add a linter for the code, do an audit of the gems we actually need, put back in place things like skylight, papertrail, etc and bring this repository back to health.

Mostly will be done once we have reviewed all pending issues and finished with the challenge edit page.

nemaniarjun:
Thinking of doing this sometime soon. 
Maybe we should use https://github.com/danmayer/coverband to start with

nemaniarjun:
We need to review models which are marked with ``??`` 

@spMohanty @skbly7 @piotrekpasciak 

# List of Models to Remove from rails
- merit/
- article.rb
- article_section.rb
- challenge_call.rb [We decided to only have one challenge call, which means we should be able to remove this model]
- comment.rb
- dataset_file_download [We have blazer]
- login_activity.rb [We have blazer]
- null_participant.rb 
- submission_comment.rb
- submission_file_definition.rb
- task_dataset_file_download.rb [We have blazer]
- topic.rb
- email_preference.rb ?? 
- email_preferences_token.rb ??
- mandrill_messange.rb ??
- notification.rb ??
- reserved_userhandle.rb ??
- views/ ??

skbly7:
I am not sure about question here, looks like lot of above file have features which we are using (like below but not limited to):

- reserved_userhandle
- notification
- email_preferences
- challenge_call?
- article (as in blog posts : https://www.aicrowd.com/blogs/publishing_aicrowd_papers)

nemaniarjun:
Let's discuss this in the call as well. Till then I'll properly try to build a case for which we should remove and why.

piotrekpasciak:
@skbly7 @nemaniarjun 
All models marked with checkbox have been removed from repository. Do we want to close this issue/

skbly7:
Yes, it seems complete now. +1 to close it. 
Great work, @piotrekpasciak, and @nemaniarjun. 🍻

PR #1077, opened 2020-01-16
Remove Topic model

piotrekpasciak:
Addressing #1002 GitHub issue


app/services/notification_service.rb
```diff
@@ -1,72 +0,0 @@
-class NotificationService
```
nemaniarjun: This should be a seperate commit.

piotrekpasciak: @nemaniarjun I've moved it to separate commit.

nemaniarjun:
LGTM!

PR #1078, opened 2020-01-16
Remove Comment model

piotrekpasciak:
Addressing #1002 GitHub issue


skbly7:
This PR include removing "comments" on Discussion side. I am currently not sure,
- Should we completely remove Discussion itself?
- Should we remove it partially now or later?

I am pretty sure we are not going to use current Discussion code, but maybe it can be reused _somehow_ to fetch Discourse things in future, say use these classes as proxy class, etc.

piotrekpasciak:
@skbly7 
I wanted to remove discussion together with Topic removal. If I correctly recall, Topic model was using it.

PR #1079, opened 2020-01-16
[WIP] CrowdAI Migration: Updated Views

nemaniarjun:
This commit updates the views and routes for the crowdai migration
will soon be changed with the right text and repushed.

@piotrekpasciak @skbly7 please review the entire controller and views, thanks! Will request on github soon


app/views/crowdai_migration/setup.html.erb
```diff
@@ -1,19 +1,86 @@
...
+<% end %>
+
 
-          <%= form_for crowdai_migration_path, method: :patch do %>
-            <%= label_tag(:data, "Encrypted Data (Received from CrowdAI):") %>
-            <%= text_field_tag(:data) %>
-            <%= submit_tag("Migrate") %>
-          <% end %>
+<% unless @current_participant %>
+  <!-- section -->
```
piotrekpasciak: You could put contents of this `unless @current_participant` and code will be claner and nothing will change on page.

app/controllers/crowdai_migration_controller.rb
```diff
@@ -21,17 +21,19 @@ def save
...
+
+                          rsa_key                   = OpenSSL::PKey::RSA.new(ENV['CROWDAI_PARTICIPANT_MIGRATION_PUBKEY'])
+                          cipherkey, iv, ciphertext = @data.split('~').map { |x| Base64.urlsafe_decode64(x) }
+                          cipher                    = OpenSSL::Cipher.new('AES-256-CBC')
+                          cipher.decrypt
+                          cipher.key = rsa_key.public_decrypt(cipherkey)
+                          cipher.iv  = iv
+                          plaintext  = cipher.update(ciphertext) + cipher.final
+                          JSON.parse(plaintext).symbolize_keys
+                        rescue StandardError => e
```
piotrekpasciak: Adding `rescue` for StandardError is very bad practice and can catch much more errors then intended. If you have time for that it would be great to list specific errors in rescue blog that we intend to catch.

app/controllers/crowdai_migration_controller.rb
```diff
@@ -21,17 +21,19 @@ def save
 
   def crowdai_params
```
piotrekpasciak: The standard approach is to put this kind of logic to separate class e.g ParseCrowdaiParamsService. Usually, we put all business logic that differs from CRUD operations in `Service Object` classes.

piotrekpasciak: Every method outside of actions in Controller should be `private`.

config/routes.rb
```diff
@@ -178,7 +178,7 @@
   match '/faq',     to: 'pages#faq',     via: :get
   match '/cookies', to: 'pages#cookies_info', via: :get
   match '/crowdai_migration', to: 'crowdai_migration#setup', via: :get
-  match '/crowdai_migration', to: 'crowdai_migration#save', via: :patch
+  match '/crowdai_migration/save', to: 'crowdai_migration#save', via: :get
```
piotrekpasciak: Your action is modifying resources on server. This action should be `POST`, as `GET` should be only used for actions that retrieve resources from server e.g rendering html template, json data.

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,11 +1,11 @@
 class CrowdaiMigrationController < ApplicationController
-  before_action :authenticate_participant!
-
-  def setup; end
+  def setup
```
piotrekpasciak: More standard approach would be to define actions called like `new` and `create` instead of `setup` and `save`.

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,11 +1,11 @@
...
+    @data = params[:data]
+    @cid  = crowdai_params[:id]
+    redirect_to(root_path, flash: { error: 'Missing data' }) if @cid.nil?
+  end
 
   def save
-    @cid = crowdai_params[:id]
-
     redirect_to(root_path, flash: { error: 'Error while processing, contact help@aicrowd.com' }) && return if @cid.nil?
```
piotrekpasciak: `return redirect_to(root_path, flash: { error: 'Error while processing, contact help@aicrowd.com' }) if @cid.nil?` would be cleaner way to write this line of code

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,11 +1,11 @@
 class CrowdaiMigrationController < ApplicationController
-  before_action :authenticate_participant!
-
-  def setup; end
+  def setup
+    @data = params[:data]
+    @cid  = crowdai_params[:id]
```
piotrekpasciak: You should use instance variables (those with @ sign) only when you need it. This should be a local variable as you seem to be using it only in next line.

nemaniarjun: I am also using it in the `save` function

piotrekpasciak: But isn't `save` a separate action and separate request to server? Wouldn't that variable always be `nil` in save action?

config/routes.rb
nemaniarjun: I put it as get as its going to be triggered by a link not by a form, so thought it was best to do it like this.

piotrekpasciak: If it has to be triggered as a link, I guess we have to go with `get`.

app/controllers/crowdai_migration_controller.rb
nemaniarjun: This seems excessive considering this is only a few lines of code, and very very specific to this use case. We probably won't be needing this code around after say --- 2 months.

piotrekpasciak: Alright, it won't be a big deal to keep it in controller.

app/controllers/crowdai_migration_controller.rb
nemaniarjun: Yeah I was confused a bit here, will modify and send in a PR soon

app/controllers/crowdai_migration_controller.rb
nemaniarjun: I think its better to go with a catch-all here, we can gracefully redirect the user, but also get error message from the logs

piotrekpasciak: But what if you make a typo in code or OpenSSL raise some kind of error we don't want to catch? Rescue block will catch it and Rollbar won't tell us there is issue in code.  We won't be able to react.

config/routes.rb
nemaniarjun: changed to post

spec/services/migrate_user_service_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+describe MigrateUserService do
+  let!(:participant) { create :participant }
```
piotrekpasciak: Standard when writing unit tests for class is to define a subject in first line within describe block e.g `subject { described_class.new(params)) }`

spec/services/migrate_user_service_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+describe MigrateUserService do
```
piotrekpasciak: If you don't have time to finish test examples, it will be better to delete test file.

nemaniarjun: I was taking a break, so didnt want to delete it, committed just as WIP :)

spec/controllers/crowdai_migration_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+describe CrowdaiMigrationController, type: :controller do
```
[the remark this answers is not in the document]
nemaniarjun: Ahh, did not know this. Will read up on them right away. :)

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,37 +1,41 @@
...
-  before_action :authenticate_participant!
+  before_action :set_crowdai_params, only: [:create]
 
-  def setup; end
-
-  def save
-    @cid = crowdai_params[:id]
+  def new
+    @data = params[:data]
+    redirect_to(root_path, flash: { error: 'Missing data' }) if @data.nil?
```
skbly7: "Missing data" will be useless to end user. Let's update this error with something useful. 

Can we also easily add blazer to record this error if possible?
This will help us in tracking non-rollbar errors while users try to migrate.

skbly7: This is kind of general comment, wherever we are exiting/redirecting from CrowdaiMigrationController to bring visibility.

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,37 +1,41 @@
...
+  end
 
-    redirect_to(root_path, flash: { error: 'Error while processing, contact help@aicrowd.com' }) && return if @cid.nil?
+  def create
+    return redirect_to(root_path, flash: { error: 'Error while processing, contact help@aicrowd.com' }) if @cid.nil?
 
     @migrate_service = MigrateUserService.new(crowdai_id: @cid, aicrowd_id: current_participant.id)
 
-    redirect_to(root_path, flash: { error: 'Participant Already Migrated' }) && return if @migrate_service.check_migrated
+    return redirect_to(root_path, flash: { error: 'Participant Already Migrated' }) if @migrate_service.check_migrated
```
skbly7: "CrowdAI account associated with this link is already migrated"

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,37 +1,41 @@
...
 
-    redirect_to(root_path, flash: { error: 'No Data exists for this user' }) && return unless MigrationMapping.exists?(crowdai_participant_id: @cid)
+    return redirect_to(root_path, flash: { error: 'No Data exists for this user, contact help@aicrowd.com' }) unless MigrationMapping.exists?(crowdai_participant_id: @cid)
```
skbly7: "No data found for your CrowdAI account, please contact help@aicrowd.com"

app/controllers/crowdai_migration_controller.rb
```diff
@@ -1,37 +1,41 @@
...
 
     MigrateUserJob.perform_later(crowdai_id: @cid, aicrowd_id: current_participant.id)
 
     redirect_to(root_path, flash: { success: 'Migration in progress.' })
```
skbly7: We have received and processing your migration request. Welcome to AIcrowd!

app/controllers/crowdai_migration_controller.rb
nemaniarjun: select * from ahoy_events where name="Processed #CrowdaiMigrationController##create"

app/services/migrate_user_service.rb
```diff
@@ -63,6 +77,21 @@ def undo_migration
         pct.update(participant_id: nil)
       end
 
+
+      votes_migration_mappings = MigrationMapping.where(crowdai_participant_id: @old_id, source_type: 'Vote')
```
piotrekpasciak: You could move code from
https://github.com/AIcrowd/AIcrowd/pull/1079/files#diff-c09e7e97bca02b16106533feabe0e7eeR81-R93
and 
https://github.com/AIcrowd/AIcrowd/pull/1079/files#diff-c09e7e97bca02b16106533feabe0e7eeR38-R51
to separate method that takes `participant_id` parameter and make this code more DRY.

nemaniarjun: I have an idea on how to make the whole thing a bit more DRY. Can you tell me where should I put the extracted method? A private method in this service itself should work right?

piotrekpasciak: Yes, private method in this service will work. Other option is to move this logic to separate class, but feel free to go with private method.

PR #1091, opened 2020-01-27
Refactor Challenge edit page 

piotrekpasciak:


app/controllers/challenges_controller.rb
```diff
@@ -120,151 +106,8 @@ def set_challenge
...
-          ],
-          invitations_attributes:                 [
-            :id,
-            :email,
-            :_destroy
-          ])
-  end
-
   def set_organizer
-    @organizer = Organizer.friendly.find(params[:organizer_id])
```
nemaniarjun: Not sure about this change. What happens when an admin opens the challenge edit page?
Admin participant wont have organizer association. 
There is no point in doing this like that. Also in the future we will want to allow participants to be part of multiple organizations.

app/policies/challenge_policy.rb
```diff
@@ -24,11 +24,11 @@ def update?
   end
 
   def new?
-    edit?
+    participant&.admin?
```
nemaniarjun: what is the reason behind this change? the `edit` function also takes care of the participants who are not admin but are organizers.

app/views/challenges/_form.html.erb
```diff
@@ -0,0 +1,51 @@
...
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-12">
+        <%= form_with model: @challenge, class: 'form-rg challenges-form', multipart: true, local: true do |f| %>
+          <p id="challenges-form-errors" class="challenges-form__errors-paragraph"><%= @challenge.errors.full_messages.to_sentence %></p>
+
+          <% hidden_field_tag :step, id: 'challenges-form-step', value: :overview %>
+
+          <div class='tab-content'>
+            <div id='tab-details' class="tab-pane fade show active">
```
nemaniarjun: Will it be better to have an array of partials here which we can loop over? Would make this code more readable and allow for changes in the future.

app/controllers/challenges_controller.rb
piotrekpasciak: That's a great catch 👍 , should be `challenge.organizer.` I've decided to drop organizer dependency for challenge.

app/policies/challenge_policy.rb
piotrekpasciak: I wanted to make sure that only admins can create new challenges, as this feature is not meant to be released.
 Previously there was no possibility to create new challenge through a form. Correct me if I'm wrong.

app/controllers/challenges_controller.rb
nemaniarjun: Yes I agree with dropping the dependency for organizer for new challenges.

app/views/challenges/_form.html.erb
piotrekpasciak: I disagree. I really feel this is so mutch easier to read thann version with loop and some kind of constants.

app/policies/challenge_policy.rb
nemaniarjun: I think there was. The organizers page (ex: https://www.aicrowd.com/organizers/imageclef) has a new challenge button which allows them to create a new challenge. The changing of the status from :draft to any other was something that only admins are allowed to do, but an organizer should be able to draft a challenge all by themselves.

app/views/challenges/_form.html.erb
nemaniarjun: Hmm, I used to think the same way, but after what happened with the challenge _subnav partial, I think we have to move to loops at some point to make things easier. 
For now of course we can leave this like it is.

app/views/challenges/_navigation.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="sub-nav-bar sticky-top">
+  <div class="container-fluid">
+    <ul class='nav'>
+      <li class="nav-item"><a class="nav-link active" data-toggle='tab' href="#tab-details" onclick="return false;">Details</a></li>
+      <% if challenge.persisted? %>
+        <li class="nav-item"><a class="nav-link" data-toggle='tab' href="#tab-overview" onclick="return false;">Overview</a></li>
+        <li class="nav-item"><a class="nav-link" data-toggle='tab' href="#tab-type" onclick="return false;">Rounds</a></li>
```
nemaniarjun: seems like a typo. Should be `tab-rounds` instead of `tab-type`

app/policies/challenge_policy.rb
piotrekpasciak: @nemaniarjun I've added commit that allows admins and organizers to create new challenge.

app/views/challenges/_form.html.erb
piotrekpasciak: I prefer to leave like that for now.

app/views/challenges/_navigation.html.erb
piotrekpasciak: Thanks, that's a good catch.

app/models/challenge.rb
```diff
@@ -73,7 +72,6 @@ class Challenge < ApplicationRecord
 
   validates :status, presence: true
   validates :challenge, presence: true
-  validates :organizer_id, presence: true
```
nemaniarjun: Hmm this seems weird. 
Every challenge has to have an organizer linked to it right ?

app/controllers/challenges_controller.rb
```diff
@@ -108,6 +110,11 @@ def set_organizer
     @organizer = @challenge&.organizer
   end
 
+  def set_organizers_for_select
+    # We'll need refactor to select2 with API endpoint when we'll have a lot of organizers.
+    @organizers_for_select = Organizer.limit(1000).pluck(:organizer, :id)
```
nemaniarjun: Organizer.all.pluck would be better right? we dont want to suddenly run into a bug due to this hardcoded number

app/models/challenge.rb
piotrekpasciak: I've deleted it, because this code is redundant, because of following line:
https://github.com/AIcrowd/AIcrowd/blob/master/app/models/challenge.rb#L8

`belongs_to :organizer` implements presence validation since Rails 5. To disable this validation on `belongs_to` you have to pass `optional: true` parameter.

Having 2 validations that do same thing has consequence of that we have 2 validation error message about same issue, what doesn't seem right.

app/controllers/challenges_controller.rb
piotrekpasciak: I've added `1000` here because  usually at point  of 1000 records it makes sense to refactor to select2 with ajax to resolve . performance issues.

I'll implement your suggestion as it's actually seems more reasonable.

app/controllers/challenges_controller.rb
```diff
@@ -55,37 +41,46 @@ def show
...
   end
 
-  def edit; end
-
   def create
-    @challenge                = @organizer.challenges.new(challenge_params)
-    @challenge.clef_challenge = true if @organizer.clef_organizer
+    @challenge                = Challenge.new(challenge_params)
+    @challenge.organizer_id   = current_participant.organizer_id if @challenge.organizer_id.blank?
+    @challenge.clef_challenge = true if @organizer&.clef_organizer
```
nemaniarjun: hmm.. if we are setting the organizer_id on line 50, then we should remove set_organizer before create and use this id for the organizer right?

piotrekpasciak: Currently this line is necessary to allow organizer to create a challenge. I'll check if I can drop it by modyfing the following input:
https://github.com/AIcrowd/AIcrowd/pull/1091/files#diff-08ced1b1da67c79e061e958b17f3ea61R147

nemaniarjun: not sure about this. 
The workflow I imagine
1. no set_organizer is called before create, its only called before new, edit and update `@organizer = @challenge&.organizer`
2. In the form using the input that you mentioned we will set the :organizer_id 
```
<% if @challenge.new_record? && current_participant&.admin? %>
      <%= f.select :organizer_id, @organizers_for_select, { include_blank: true } %>
    <% else %>
      <input class="form-control" type='text' value='<%= @challenge.organizer&.organizer || current_participant.organizer&.organizer %>' readonly>
    <% end %>
```
3. In create we use this id that was set to :organizer_id and use that for line 51
`@challenge.clef_challenge = true if @challenge.organizer&.clef_organizer`
or 
`@challenge.clef_challenge = true if Organizer.find(@challenge.organizer_id)&.clef_organizer`

nemaniarjun: Ah I just noticed that the readonly is the text part. we should also have an organizer_id field there which is readonly and set using `@challenge.organizer&.id || current_participant.organizer&.id`

nemaniarjun: On third thought, why are we setting challenge.clef_challenge to true? maybe we should make it a model attribute which just calls self.organizer.clef_organizer

piotrekpasciak: https://github.com/AIcrowd/AIcrowd/pull/1091#discussion_r371818837

piotrekpasciak: https://github.com/AIcrowd/AIcrowd/pull/1091#discussion_r371819475
I don't know why.  This is old code and I'm no familiar with business logic. Decided to not touch it.

nemaniarjun: Another idea is to have a select field for when an participant creates the challenge with a select field for all the organizations that the participant is a part of. 
We want to change the one-to-one mapping of participant as organizer to multiple allowing a participant to be a part of multiple organization

piotrekpasciak: https://github.com/AIcrowd/AIcrowd/pull/1091#discussion_r371817874 I've implement this remark.

piotrekpasciak: https://github.com/AIcrowd/AIcrowd/pull/1091#discussion_r371821357 - I think this is additional work. Don't want to add additional work to this PR. If this is needed I prefer to due it on separate branch.

app/controllers/challenges_controller.rb
```diff
@@ -55,37 +41,45 @@ def show
   end
 
   def new
-    @challenge = @organizer.challenges.new
+    @challenge = Challenge.new(organizer: @organizer)
```
nemaniarjun: is this @organizer set now? :thinking: :thought_balloon:

piotrekpasciak: @nemaniarjun  This is not needed anymore. I'll remove it.

piotrekpasciak: @organizer is no longer set after previous changes

app/policies/challenge_policy.rb
```diff
@@ -24,7 +24,7 @@ def update?
   end
 
   def new?
-    edit?
+    participant && (participant.admin? || @participant.organizer_id.present?)
```
nemaniarjun: is it `participant` or `@participant`?

piotrekpasciak: Both point to the same thing, I'll update it.

app/assets/stylesheets/challenges/form.scss
```diff
@@ -0,0 +1,11 @@
+.challenges-form {
```
nemaniarjun: Okay, one last thing. Can you move these css changes to a separate commit? that will help later when we work with jason to update the styles. its not required but having a different commit will probably work in our favour for finding and changing this later.

piotrekpasciak: I've moved css styles to separate commit.

app/views/challenges/form/_form.html.erb
```diff
@@ -1,78 +0,0 @@
...
-          <div id='tab-overview' class="tab-pane fade">
-            <%= render partial: 'challenges/form/overview', locals: { challenge: challenge } %>
-          </div>
-
-          <div id='tab-private' class="tab-pane fade">
-            <%= render partial: 'challenges/form/private_challenge', locals: { challenge: challenge } %>
-          </div>
-
-          <div id='tab-dataset' class="tab-pane fade">
-            <%= render partial: 'challenges/form/dataset', locals: { challenge: challenge } %>
```
nemaniarjun: @piotrekpasciak

piotrekpasciak: @nemaniarjun After refactor we no longer render it.

PR #1092, opened 2020-01-27
Show next tab after updating the challenge

piotrekpasciak:


app/views/challenges/_navigation.html.erb
```diff
@@ -1,16 +1,16 @@
 <div class="sub-nav-bar sticky-top">
   <div class="container-fluid">
     <ul class='nav'>
-      <li class="nav-item"><a class="nav-link active" data-toggle='tab' href="#tab-details" onclick="return false;">Details</a></li>
+      <li class="nav-item"><a id="challenge-edit-details-tab" class="nav-link active" data-toggle='tab' href="#tab-details" onclick="return false;">Details</a></li>
```
nemaniarjun: Can you check if the onclick="return false" is still needed or not?

app/views/challenges/tabs/_admin_form.html.erb
```diff
@@ -22,5 +22,5 @@
     <%= f.text_field :dynamic_content_url, class: "form-control" %>
   </div>
 
-  <%= render partial: 'challenges/fields/submit_buttons', locals: { f: f } %>
+  <%= render partial: 'challenges/fields/submit_buttons', locals: { f: f, next_step: 'admin' } %>
```
nemaniarjun: Instead of hard coding the next step in the params, should we pass the next step with the step or in some other way that is more direct?

piotrekpasciak: `should we pass the next step with the step` - I don't understand this approach

There are many approaches e.g: hardcoding  steps and their order on back-end, doing that on front-end

app/views/challenges/_navigation.html.erb
piotrekpasciak: I've been testing a little bit and it seems that this is not needed. Good catch 👍

app/views/challenges/tabs/_admin_form.html.erb
nemaniarjun: This seems straightforward, if we are making more changes to the form later we can consider refactoring this later. lets leave this for now.

app/javascript/stimulus_js/controllers/challenges_edit_controller.js
```diff
@@ -0,0 +1,15 @@
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
+  initialize() {
+    const currentUrl  = new URL(window.location);
+    const currentStep = currentUrl.searchParams.get('step');
+    // I have to use JQuery here, because Bootstrap tabs use JQuery and
+    // plain javascript won't trigger events on them
```
nemaniarjun: no need for this comment. we should use jqeury where possible anyway.

piotrekpasciak: Why do . you think that? ES6 implements all standard funcionalities of JQuery? I think JQuery is legacy approach and we would want to drop it if it were possible without mutch work.

nemaniarjun: I am not so sure about ES6 having all functionality of jquery. 
Also why cant we trigger events using plain js ? I dont understand that part well. 

My point was if we are dependent on jquery for some use cases we should go all in and use that wherever we can or we should figure out a why normal es6 is not going to trigger these events.

nemaniarjun: Anyway, lets leave this comment here for now, we can come back to this in future sprint when we refactor and remove all the js that we have to use webpack instead of paloma / standard rails pipeline.

piotrekpasciak: `Also why cant we trigger events using plain js ?` -  Javascript functions are not available when I select the node with ES6, while they're available when I select the node with JQuery.

piotrekpasciak:
@nemaniarjun @skbly7 
I would prefer to freeze it for now, as I need to finish ajax submit first.

piotrekpasciak:
@skbly7 I acutally didn't want to merge it yet.

PR #1093, opened 2020-01-27
Fix missing client side validations for Challenge edit form

piotrekpasciak:
I'll add Image_file validations in separate PR.


app/views/challenges/fields/_challenge_rules_fields.html.erb
```diff
@@ -24,3 +24,8 @@
     <%= f.text_area :additional_checkbox_text_markdown, class: "form-control txt-sml" %>
   <% end %>
 </div>
+
+<script>
+  // I have to refresh client_side_validations after adding the association
+  $('#challenges-form').enableClientSideValidations();
```
skbly7: Is it possible to add it at a single place so we don't forget for new tabs in future? Maybe in challenge tab animation js code?

piotrekpasciak: This has to be in this file, as if I move it elsewhere new nested forms . won't handle the validation.

piotrekpasciak: https://github.com/AIcrowd/AIcrowd/blob/master/app/views/challenges/tabs/_rules_form.html.erb#L12

This link won't work correctly if I move it outside this partial.

nemaniarjun:
You can check the PR for challenge rounds. I did something similar there
using a cocoon callback, maybe you can use that too.

piotrekpasciak:
@nemaniarjun 
I think approach you're proposing won't work because this javascript line needs to be executed after association partial gets inserted to html.

If you want to proof I'm wrong go ahead and show me . working solution.

app/views/challenges/fields/_challenge_rules_fields.html.erb
skbly7: Callback addition somewhere like here with 1000 delay: https://github.com/AIcrowd/AIcrowd/blob/45f339b1af0a69f9576c94596a55dc3418887508/app/assets/javascripts/controllers/challenges_controller.js#L38

piotrekpasciak: Screenshot 2020-01-28 at 18 31 56

@skbly7 Did you mean something like that?

piotrekpasciak: @skbly7 It doesn't work for me. Are you sure this should work? Can we jump . on a call where I'll show you the edge case?

piotrekpasciak: @skbly7  Showed me how to make it work properly.

piotrekpasciak: I've updated the code to include your remark.

skbly7: It looks great now!

PR #1096, opened 2020-01-27
Refactor: Radio Buttons

nemaniarjun:
There are 2 commits, one which refactors the use of the styled radio button on the leaderboards and submissions pages, and the other which creates a nice partial for the radio button and changes all checkboxes in the challenge edit page to use the new partial.


app/controllers/leaderboards_controller.rb
```diff
@@ -6,13 +6,8 @@ class LeaderboardsController < ApplicationController
...
   def index
     @current_round = current_round
-    if @challenge.completed?
-      @post_challenge = if params[:post_challenge] == 'on'
-                          'on'
-                        else
-                          'off'
-                        end
-    end
+    @post_challenge = true if @challenge.completed? && params[:post_challenge].presence == "true"
```
piotrekpasciak: `params[:post_challenge].presence` - `presence` method is redundant here. It doesn't have any impact on this logical condition.

app/views/leaderboards/index.html.erb
```diff
@@ -42,44 +42,21 @@
...
-				      <% end %>
-			      </div>
-		      </div>
-	      <% end %>
-      <!-- /toggle switch -->
+        <!-- toggle switch -->
+        <% if @challenge.status == :completed %>
+          <div class="toggle-switch-container">
+            <div class="can-toggle can-toggle--size-small">
+              <%= link_to challenge_leaderboards_path(challenge_round_id: @current_round, post_challenge: !@post_challenge) do %>
```
piotrekpasciak: What's the purpose of `!@post_challenge`? Why do you have `!` operator here? Did you mean something like `@post_challenge.blank?`

app/views/submissions/index.html.erb
```diff
@@ -49,31 +49,32 @@
...
-          <%= render partial: 'submissions/index/baselines_off',
-                locals: { challenge: @challenge, current_round_id: @current_round_id } %>
-          <% end %>
-          <% if @baselines == 'on' %>
-          <%= render partial: 'submissions/index/baselines_on',
-                locals: { challenge: @challenge, current_round_id: @current_round_id } %>
-          <% end %>
+          <div class="toggle-switch-container">
+            <div class="can-toggle can-toggle--size-small">
+              <%= link_to challenge_submissions_path(@challenge, my_submissions: !@my_submissions, current_round_id: @current_round_id) do %>
```
piotrekpasciak: `!@my_submissions` - same questions like above.

app/views/submissions/index.html.erb
```diff
@@ -49,31 +49,32 @@
...
+                <label for="my-submissions">
+                  <div class="can-toggle__switch" data-checked="On" data-unchecked="Off"></div>
+                  <div class="can-toggle__label-text">My Submissions</div>
+                </label>
+              <% end %>
+            </div>
+          </div>
+          <div class="toggle-switch-container">
+            <div class="can-toggle can-toggle--size-small">
+              <%= link_to challenge_submissions_path(@challenge, baselines: !@baselines, current_round_id: @current_round_id) do %>
```
piotrekpasciak: `!baselines`

app/views/leaderboards/index.html.erb
nemaniarjun: if @post_challenge is `nil` we want to pass `true` here if its `true` we want to pass `false` here

app/controllers/leaderboards_controller.rb
nemaniarjun: aah, will change this!

app/views/leaderboards/index.html.erb
piotrekpasciak: Can we use `@post_challenge.blank?` in that case? It's hard to read code with `!` operators.

nemaniarjun: initally I was setting @post_challenge to `false` that is why I was using `!` but now in the controller we are only setting it if its true so yes I will change it to `.blank` that is definitely more readable

PR #1103, opened 2020-01-30
Add primary_sort_order, secondary_sort_order, score_title and score_secondary_title to challenge_rounds

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/367862217

Tasks to run after the deploy:
`bundle exec rake challenges:migrate_score_titles_to_rounds`


app/concepts/leaderboard/views/table_head.erb
```diff
@@ -6,9 +6,9 @@
     <% if challenge.media_on_leaderboard %>
         <th scope="col">Media</th>
     <% end %>
-    <th scope="col"><%= challenge.score_title %></th>
```
skbly7: Hi, due to legacy issue and external services I would like to stick to `score` and `score_secondary` naming instead of `primary_score` and `secondary_score`. 
The external services doesn't fill in `_title` fields but they update database for `score` & `score_secondary`, which may cause confusion in future.

piotrekpasciak: But will these external services work when these fields are in different database table?

skbly7: Yes, they update `score` and `score_secondary` fields in Submission model.

lib/tasks/challenges.rake
```diff
@@ -0,0 +1,16 @@
+namespace :challenges do
+  desc "Migrate challenges score titles to rounds score titles"
+  task migrate_score_titles_to_rounds: :environment do
+    Challenge.includes(:challenge_rounds).find_each do |challenge|
+      primary_score_title   = challenge.score_title
```
skbly7: Maybe add an if condition in case titles aren't set, we have lot of challenges with these values as blank, would migration run just fine in that case?

app/concepts/leaderboard/views/table_head.erb
piotrekpasciak: Allright, I'll update that after standup.

lib/tasks/challenges.rake
piotrekpasciak: Migration will run just fine, but will be a little bit faster if I add these guard clauses.

piotrekpasciak: I've added guard clause as you've proposed.

app/views/challenges/tabs/_details_form.html.erb
```diff
@@ -93,24 +93,12 @@
 
   <div class="form-row">
     <div class="form-group col-6">
-      <label for="primaryScoreTitle">Primary score title</label>
-      <%= f.text_field :score_title, class: "form-control" %>
-    </div>
-
-    <div class="form-group col-6">
-      <label for="primaryScoreTitleSortOrder">Sort order</label>
+      <label for="primaryScoreTitleSortOrder">Primary score sort order</label>
```
skbly7: Hi, I am sorry I didn't realise during first pass review.

We have left `primary_sort_order` and `secondary_sort_order` in the challenge, but it make sense to move it to challenge round setting as well.

piotrekpasciak: @skbly7 What do you think about doing it in separate PR? This is significant amount of work.

skbly7: @piotrekpasciak I will stick with these changes to go together in this & #1104, reason being scores module as a whole is usable only when name+sort order together can be customised. 

Plus, the migrations etc would make sense to remain together in both PRs, because of similarity.

Issue #1065, opened 2020-01-14
Discourse related error handling


skbly7:
Due to the nature of how Discourse is added in our rails codebase, it ends up having 500 error for multiple reasons, which include:

- category doesn't exist
- category name is too large for discourse to handle (as we have no limit on competition length or having longer length allowed)
- discourse is down (hypothetically it will cause 500 as well)
- etc..

We need proper fix for this. Worst case/Atleast, try/catch block which hide the discussion link on challenges page when discourse is causing any error.

piotrekpasciak:
@skbly7 Fix for this issue is already on staging.

nemaniarjun:
Closing issue as already completed.

PR #1106, opened 2020-01-31
Refactor Discourse integration to ServiceObject classes

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/429654890

Github Issue:
https://github.com/AIcrowd/AIcrowd/issues/1065

Discourse Network errors are handled in `discourse_api` gem:
https://github.com/discourse/discourse_api/blob/master/lib/discourse_api/client.rb#L143-L144

Change of uniqueness collision is 1 to (256^6 - Challenge.count), thus I've decided to ignore it.


app/services/discourse/create_category_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      response = create_category_request(ensure_uniqueness: retry_count.positive?)
+      challenge.update!(discourse_category_id: response['id'])
+
+      success
+    rescue DiscourseApi::UnprocessableEntity => e
+      raise e if retry_count.positive?
+
+      retry_count += 1
+
+      retry if e.message.include?('Category Name has already been taken')
```
skbly7: That's nice to have random string only when needed! Well thought implementation. 💯

app/services/discourse/update_category_service.rb
```diff
@@ -0,0 +1,48 @@
+module Discourse
+  class UpdateCategoryService < ::Discourse::BaseService
+    def initialize(challenge:)
+      @client    = prepare_api_client
+      @challenge = challenge
+    end
+
+    def call
```
skbly7: (Minor) This function is similar to `create_cateogry_service -> call` and can be potentially refactored.

Some function say `def update_of_create` in Discourse::BaseService which knows to send create request v/s update based on parameters

spec/services/discourse/create_category_service_spec.rb
```diff
@@ -0,0 +1,82 @@
+require 'rails_helper'
```
skbly7: 💯

app/services/discourse/update_category_service.rb
piotrekpasciak: That kind of action is usually called `upsert`. If we decide to implement it, I would simply merge both services into `UpsertService` class.

skbly7: Sure. `UpsertService` would be good fit for this usecase.

PR #1111, opened 2020-02-04
Refactor Votes cells to partials

piotrekpasciak:


app/views/shared/challenges/_masthead.html.erb
```diff
@@ -76,7 +76,11 @@
           </div>
         </div>
         <div class="btn-set" aria-hidden="true">
-          <%= concept(Vote::Cell, challenge, current_participant: current_participant) %>
+          <%
+            # TODO: Move this query to controller
+            vote = current_participant.present? ? challenge.votes.where(participant_id: current_participant&.id).first : nil
```
nemaniarjun: What is the blocker with doing this right away? seems like this should be called on all challenge pages

piotrekpasciak: I've moved this database query to controllers.

PR #1118, opened 2020-02-05
Refactor Submissions cells to partials

piotrekpasciak:


nemaniarjun:
Kind of worried about the diff coverage? We should add tests for the helpers?

piotrekpasciak:
@nemaniarjun Test coverage didn't drop at all due to this PR. I don't want to put additional work here to add test coverage as additional refactor will be needed after we migrate out of Traiblazer Cells.

Traiblazer cells have a lot of duplicated code what you will notice in next refactor PRs.

PR #1123, opened 2020-02-07
Remove useless materialized views

piotrekpasciak:


app/policies/leaderboard_policy.rb
```diff
@@ -76,6 +77,7 @@ def resolve
           ]
           scope.where(sql)
         else
+          binding.pry
```
nemaniarjun: Do we need these debugging statements ?

piotrekpasciak: Definitely not, I forgot to remove them. PR is in work-in-progress state anyway.

nemaniarjun:
Once https://github.com/AIcrowd/AIcrowd/pull/1126 is merged we can also remove the challenge_round_summary view.

nemaniarjun:
Now that #1126 is merged, we are not using the challenge_round_summary view anywhere in the codebase. we should remove that view as well in this PR. :smile:

PR #1126, opened 2020-02-09
WIP: Add defaults to challenge edit page

nemaniarjun:


app/concepts/leaderboard/cells/challenge_round_pills.rb
```diff
@@ -12,7 +12,7 @@ def current_round
   end
 
   def challenge_rounds
-    challenge.challenge_round_summaries.where(round_status_cd: ['history', 'current'])
+    challenge.started_rounds
```
piotrekpasciak: 👍

piotrekpasciak: Are you sure this should have `started_rounds` scope? This query seems to be different than `.challenge_rounds.where("start_dttm < ?", Time.current)`

app/controllers/challenges_controller.rb
```diff
@@ -33,12 +33,12 @@ def show
...
       @challenge_participant = @challenge.challenge_participants.where(
         participant_id:                   current_participant.id,
-        challenge_rules_accepted_version: @challenge.current_challenge_rules_version)
+        challenge_rules_accepted_version: @challenge.current_challenge_rules&.version)
     end
 
     @challenge.record_page_view unless params[:version] # dont' record page views on history pages
     @challenge_rules  = @challenge.current_challenge_rules
-    @challenge_rounds = @challenge.challenge_rounds.where("start_dttm < ?", Time.current)
+    @challenge_rounds = @challenge.started_rounds
```
piotrekpasciak: 👍

app/models/challenge.rb
```diff
@@ -120,36 +126,41 @@ def status_formatted
...
     SubmissionsRemainingQuery.new(challenge: self, participant_id: participant_id).call
   end
 
-  def current_round
-    @current_round ||= challenge_round_summaries.where(round_status_cd: 'current').first
+  def active_round
+    @active_round ||= challenge_rounds.find_by(active: true)
+  end
+
+  def started_rounds
```
piotrekpasciak: This should be a scope in ChallengeRound model:
https://devblast.com/b/jutsu-11-rails-scopes

app/views/leaderboards/index.html.erb
```diff
@@ -8,10 +8,10 @@
       <%= concept(Leaderboard::Cell::ChallengeRoundPills, @challenge, current_round: @current_round) %>
 
       <!-- challenge leaderboard -->
-        <% if @current_round.leaderboard_note.present? %>
+        <% if @current_round&.leaderboard_note.present? %>
```
piotrekpasciak: You seem to be using `@active_round` in other places. Are you sure it should be called `@current_round` here?

app/concepts/leaderboard/cells/challenge_round_pills.rb
nemaniarjun: @piotrekpasciak I want to remove the `challenge_round_summaries` view from the app. 
The `challenge_rounds` calculated here actually should show all rounds that are started (thus [history, current])

app/views/leaderboards/index.html.erb
nemaniarjun: current_round -> for submissions and leaderboard filtering.
active_round -> The round which is currently active for the challenge.

app/models/challenge.rb
nemaniarjun: adding scope `started` to ChallengeRound Model 
usage -> `@challenge.challenge_rounds.started`

app/controllers/leaderboards_controller.rb
```diff
@@ -55,9 +54,9 @@ def set_follow
 
   def current_round
     if params[:challenge_round_id].present?
-      ChallengeRound.find(params[:challenge_round_id].to_i)
+      @challenge.challenge_rounds.find(params[:challenge_round_id].to_i)
```
nemaniarjun: Changed this to a scoped query over only this challenge.

app/models/challenge.rb
```diff
@@ -86,15 +86,21 @@ class Challenge < ApplicationRecord
...
       self.challenge_client_name ||= "challenge_#{SecureRandom.hex}"
-      self.featured_sequence       = Challenge.count + 1
+      self.featured_sequence     ||= Challenge.count + 1
+      self.team_freeze_time      ||= Time.now.utc + 2.months + 3.weeks
     end
   end
 
+  def create_default_associations
+    ChallengeRound.create(challenge: self)
+    ChallengeRules.create(challenge: self)
```
piotrekpasciak: `create!` - and please doublecheck if it works.

app/models/challenge.rb
```diff
@@ -86,15 +86,21 @@ class Challenge < ApplicationRecord
...
-      self.submission_license    ||= 'Please upload your submissions and include a detailed description of the methodology, techniques and insights leveraged with this submission. After the end of the challenge, these comments will be made public, and the submitted code and models will be freely available to other AIcrowd participants. All submitted content will be licensed under Creative Commons (CC).'
       self.challenge_client_name ||= "challenge_#{SecureRandom.hex}"
-      self.featured_sequence       = Challenge.count + 1
+      self.featured_sequence     ||= Challenge.count + 1
+      self.team_freeze_time      ||= Time.now.utc + 2.months + 3.weeks
     end
   end
 
+  def create_default_associations
+    ChallengeRound.create(challenge: self)
```
piotrekpasciak: We should follow policy of failing early. You should use `create!` here. Additionally, I think it's very likely this code doesn't work.  Challenge Round require more fields then` challenge_id` to be created.

app/models/challenge_rules.rb
```diff
@@ -3,10 +3,6 @@ class ChallengeRules < ApplicationRecord
   belongs_to :challenge
 
   before_create do
-    self.version = if challenge.current_challenge_rules
-                     challenge.current_challenge_rules.version + 1
-                   else
-                     1
-                   end
+    self.version = (challenge&.current_challenge_rules&.version || 0) + 1
```
piotrekpasciak: Using `version.to_i` should be cleaner then `|| 0`

app/queries/submissions_remaining_query.rb
```diff
@@ -11,9 +11,7 @@ def initialize(challenge:, participant_id:)
 
   def call
     if @challenge.running?
-      remaining = send(@challenge
-        .current_round
-        .submission_limit_period_cd.to_s)
+      remaining = send(@challenge.active_round.submission_limit_period_cd.to_s)
```
piotrekpasciak: You should avoid using `send` method at all costs, as it can call private methods. If you really need to use meta programming use `public_send` method.

app/views/leaderboards/index.html.erb
```diff
@@ -19,11 +19,11 @@
...
         <% end %>
 
-        <!-- challenge leaderboard -->
-        <% if @current_round.leaderboard_note.present? %>
+      <!-- challenge leaderboard -->
+        <% if @current_round&.leaderboard_note.present? %>
 		      <div class="md-content width-100">
 			      <div class="note">
-				      <%= sanitize(@current_round.leaderboard_note) %>
+				      <%= sanitize(@current_round&.leaderboard_note) %>
```
piotrekpasciak: `&` operator is not needed here. You check the presenece in `if` statement above,

spec/controllers/challenges_controller_spec.rb
```diff
@@ -15,7 +15,7 @@
     describe 'GET #index challenge_running' do
       before { get :index }
 
-      it { expect(assigns(:challenges)).to eq [challenge_running, challenge_completed, challenge_starting_soon] }
+      it { expect(assigns(:challenges)).to contain_exactly(challenge_running, challenge_completed, challenge_starting_soon) }
```
piotrekpasciak: Why did you change to `contain_exactly` here? Was there some kind of issue with ordering?

app/queries/submissions_remaining_query.rb
nemaniarjun: I didn't write this query, just changed the whitespace. I will drop this change if needed for now.

app/models/challenge_rules.rb
nemaniarjun: :+1:

app/models/challenge.rb
nemaniarjun: As per my testing it looks like the challenge round is created without any additional params needed.

app/models/challenge.rb
nemaniarjun: Looks like it works. If you also see the changes made to the tests, I have removed creation of ChallengeRules from the setup of specs.

app/views/leaderboards/index.html.erb
nemaniarjun: :+1:

spec/controllers/challenges_controller_spec.rb
nemaniarjun: I am not sure why I need to make this change now. The spec fails without this change
Expected: Array of challenges, Got:  ActiveRecord:Relation.

See: https://stackoverflow.com/questions/47538214/why-do-i-get-active-record-relation-instead-of-array-in-rspec

app/models/challenge.rb
piotrekpasciak: You're right. I wasn't aware that you've added defaults everywhere and thanks them all validations pass.

app/queries/submissions_remaining_query.rb
piotrekpasciak: I think you should change this line to `public_send(@challenge.active_round.submission_limit_period_cd.to_s)` anyway. We have good coverage for this class and change has low risk of failure.

nemaniarjun: I have added a commit trying to simplify this whole file a bit :)

piotrekpasciak: Your change looks mutch better 👍

PR #1128, opened 2020-02-10
Refactor ParticipantClefTask cells to partials

piotrekpasciak:


app/views/task_dataset_files/index.html.erb
```diff
@@ -9,10 +9,17 @@
...
+
+            <% if participant_clef_task_status == 'profile_incomplete' %>
+              <%= render partial: 'task_dataset_files/resources/profile_incomplete', locals: { organizer: @clef_task.organizer } %>
+            <% elsif participant_clef_task_status == 'unregistered' || participant_clef_task_status == 'requested' %>
+              <%= render partial: 'task_dataset_files/resources/participant_unregistered', locals: { clef_task: @clef_task, challenge_id: @challenge.id } %>
+            <% elsif participant_clef_task_status == 'submitted' %>
+              <%= render partial: 'task_dataset_files/resources/participant_submitted', locals: { clef_task: @clef_task, participant_clef_task: @current_participant_clef_task, challenge_id: @challenge.id } %>
+            <% elsif participant_clef_task_status == 'registered' %>
+              <%= render partial: 'task_dataset_files/resources/task_dataset_files', locals: { clef_task: @clef_task, challenge_id: @challenge.id } %>
+            <% end %>
```
alkeshrails: I thought instead of nested if-else, we should use switch case in a view helper.

config/routes.rb
```diff
@@ -132,7 +132,8 @@
       get :submission_detail
     end
     get :remove_image, on: :member
-    get :clef_task
+    get :clef_task, on: :member
+
```
alkeshrails: We can use member block here, we have two member methods 'remove_image' and 'clef_task'

app/views/task_dataset_files/index.html.erb
piotrekpasciak: i've moved code to `case` block

config/routes.rb
piotrekpasciak: I've changed code to member block.

PR #1130, opened 2020-02-11
Move methods that are not used outside of class to private section in app/controllers folder

alkeshrails:
Use private section for methods that are not used outside of class.
link : https://aicrowd.monday.com/boards/363093650/pulses/420516667

### DIFF
--- a/app/controllers/api/external_graders_controller.rb
+++ b/app/controllers/api/external_graders_controller.rb
@@ -218,11 +218,6 @@ class Api::ExternalGradersController < Api::BaseController
     end
   end
 
-  # Unused method
-  def validate_s3_key(s3_key)
-    S3Service.new(s3_key, shared_bucket = true).valid_key?
-  end
-
   class DeveloperAPIKeyInvalid < StandardError
     def initialize(msg = "The API key did not match any participant record.")
       super
--- a/app/controllers/api/external_graders_controller.rb
+++ b/app/controllers/api/external_graders_controller.rb
@@ -218,7 +218,7 @@ class Api::ExternalGradersController < Api::BaseController
     end
   end
 
-  # didn't call from anywhere
+  # Unused method
   def validate_s3_key(s3_key)
     S3Service.new(s3_key, shared_bucket = true).valid_key?
   end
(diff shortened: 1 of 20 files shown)


app/controllers/api/external_graders_controller.rb
```diff
@@ -245,105 +218,11 @@ def post_challenge(challenge, params)
...
-    return false if participant_leaderboard.nil?
-    if participant_leaderboard.score >= min_score
-      return true
-    else
-      return false
-    end
-  end
-
+  # Unused method
   def validate_s3_key(s3_key)
```
piotrekpasciak: Please remove that method.

app/controllers/challenge_rules_controller.rb
```diff
@@ -6,6 +6,7 @@ def show; end
 
   private
 
+  # Unused method
   def challenge_rules_params
```
piotrekpasciak: Just remove the method.

app/controllers/participation_terms_controller.rb
```diff
@@ -8,6 +8,7 @@ def show
 
   private
 
+  # Unused method
   def participation_terms_params
```
piotrekpasciak: Remove it.

app/controllers/api/old_submissions_controller.rb
```diff
@@ -39,6 +39,7 @@ def index
   private
 
   def set_submissions(challenge_id, grading_status, after, challenge_round_id)
+    # can manage it in multi level if-else loop
```
piotrekpasciak: You're right this method seems to be major overkill. You can leave it  for now, but if you really want to refactor it, do it in separate PR.

piotrekpasciak:
@alkeshrails 
Could you also take a look on Codeclimate issues? Would it be easy to fix them?
https://codeclimate.com/github/AIcrowd/AIcrowd/pull/1130

app/controllers/api/old_submissions_controller.rb
alkeshrails: Okay I will handle it in a separate PR. creating ticket for this.

alkeshrails:
This PR is created by fork of repo. I create a new PR for same to app repo.

Issue #788, opened 2019-01-16
Challenge Export functionality


spMohanty:
We need a functionality to "export" a challenge, and then easily "import" the said challenge.
This will both help in cloning challenges in the future, and also to help interested challenge organizers (from crowdAI) to migrate their challenge content to AIcrowd

As a first step, we simply focus on the challenge specific content, and not anything that has a reference to the individual user.

skbly7:
Lets focus on only database related things in this export/import functionality on first go (i.e. no image, etc blobs) which can form `json` file easily. (i.e. text, boolean, and similar information types)

This include to include:
- challenge model.
- challenge rounds model.
- rules model.
- partner model. (extended)
- any other association with challenge model (extended)

skbly7:
@nemaniarjun can check in case any of migration export script can help our purpose here

skbly7:
Export/Import buttons will be added in Admin tab

piotrekpasciak:
Challenge associations:
Challenge Import:

Challenge Associations (association_name => table_name):
slugs => friendly_id_slugs
organizer => organizers
clef_task => clef_tasks
dataset_files => dataset_files
submission_file_definitions => submission_file_definitions
challenge_partners => challenge_partners
challenge_rules => challenge_rules
challenge_participants => challenge_participants
submissions => submissions
votes => votes
follows => follows
challenge_rounds => challenge_rounds
invitations => invitations
teams => teams
leaderboards => leaderboards (materialised view)
ongoing_leaderboards => ongoing_leaderboards (materialised view)
participant_challenges => participant_challenges (materialised view)
participant_challenge_counts => participant_challenge_counts (materialised view)
challenge_registrations => challenge_registrations (materialised view)
challenge_organizer_participants => challenge_organizer_participants (materialised view)
challenge_round_summaries => challenge_round_summaries (materialised view)

piotrekpasciak:
Challenge fields:
    t.integer "organizer_id"
    t.string "challenge"
    t.string "status_cd", default: "draft"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.string "tagline"
    t.string "primary_sort_order_cd", default: "ascending"
    t.string "secondary_sort_order_cd", default: "not_used"
    t.boolean "perpetual_challenge", default: false
    t.string "answer_file_s3_key"
    t.integer "page_views", default: 0
    t.integer "participant_count", default: 0
    t.integer "submissions_count", default: 0
[...]

PR #1134, opened 2020-02-11
Add Challenge exporting to JSON functionality

piotrekpasciak:
Monday ticket:
https://aicrowd.monday.com/boards/363093650/pulses/447286098

Github Issue:
https://github.com/AIcrowd/AIcrowd/issues/788

Blocked until questions in Github Issue gest answered.

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -2,6 +2,89 @@ class Challenge < ApplicationRecord
   include FriendlyId
   include Markdownable
 
+  IMPORTABLE_FIELDS = [
+    :challenge,
+    :status_cd,
+    :tagline,
+    :primary_sort_order_cd,
+    :secondary_sort_order_cd,
+    :perpetual_challenge,
+    :answer_file_s3_key,
+    :score_title,
+    :score_secondary_title,
+    :slug,
+    :submission_license,
+    :api_required,
+    :media_on_leaderboard,
+    :challenge_client_name,
+    :online_grading,
+    :description_markdown,
+    :description,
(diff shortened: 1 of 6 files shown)


spec/requests/challenges_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      it 'sends json file in response' do
+        get export_challenge_path(challenge)
+
+        expect(response).to have_http_status :ok
+        expect(response.header['Content-Type']).to eq 'application/json'
+
+        challenge_data = JSON.parse(response.body)
+
+        expect(challenge_data['challenge']).to eq 'Challenge Title'
```
nemaniarjun: 💯

app/models/challenge.rb
```diff
@@ -40,8 +123,6 @@ class Challenge < ApplicationRecord
            class_name: 'ParticipantChallengeCount'
   has_many :challenge_registrations,
            class_name: 'ChallengeRegistration'
-  has_many :challenge_organizer_participants,
```
alkeshrails: Why we require to remove this association?

piotrekpasciak: This relation is broken, and was causing issues for me. I'll double check if I can make it work easly.

piotrekpasciak: @alkeshrails 
ChallengeOrganizerParticipant doesn't have `challenge_id` foreign key. This association simply cannot work.
https://github.com/AIcrowd/AIcrowd/blob/master/db/schema.rb#L990

PR #1135, opened 2020-02-12
Replace Image-cell to image_tag

alkeshrails:
link : https://aicrowd.monday.com/boards/363093650/pulses/456260224


piotrekpasciak:
@alkeshrails Code looks good, but you need to handle specs.

PR #1141, opened 2020-02-17
Challenge Subnav: Priority Nav Refactor

nemaniarjun:


app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -1,3 +1,62 @@
...
+    init: function() {
+        this.menu = $(".priority-nav");
+        this.allNavElements = $(".priority-nav > ul > li:not('.overflow-nav')");
+        this.bindUIActions();
+        this.setupMenu();
+    },
+
+    setupMenu: function() {
+        let firstPos = $(".priority-nav > ul > li:first").position();
+        let wrappedElements = $();
```
piotrekpasciak: Are you sure this line will work? Can you do something like `$()`? What will it do?

app/helpers/challenges_helper.rb
```diff
@@ -12,27 +12,17 @@ def status_dropdown_helper(builder:)
...
     if !policy(challenge).has_accepted_participation_terms?
-      return url_for([challenge, ParticipationTerms.current_terms])
+      [challenge, ParticipationTerms.current_terms]
     elsif !policy(challenge).has_accepted_challenge_rules?
-      return url_for([challenge, challenge.current_challenge_rules])
+      [challenge, challenge.current_challenge_rules]
     end
 
-    return nil
+    nil
```
piotrekpasciak: `nil` here is redundant. Methods in Ruby return nil by default.

piotrekpasciak:
Looks like some specs don't pass, please take a look on them.

app/assets/javascripts/controllers/challenges_controller.js
nemaniarjun: It creates an empty set / array for jquery to add elements too

nemaniarjun:
> Looks like some specs don't pass, please take a look on them.

will fix the specs today!

PR #1142, opened 2020-02-17
Add functionality to import Challenge from json

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/458100646

Github issue:
https://github.com/AIcrowd/AIcrowd/issues/788


app/controllers/challenges_controller.rb
```diff
@@ -107,13 +107,23 @@ def remove_image
...
               filename: "#{@challenge.challenge.to_s.parameterize.underscore}_export.json"
   end
 
+  def import
+    result = Challenges::ImportService.new(import_file: params[:import_file], organizer: @challenge.organizer).call
+
+    if result.success?
+      redirect_to edit_challenge_path(@challenge, step: :admin), notice: 'New Challenge Imported'
+    else
+      redirect_to edit_challenge_path(@challenge, step: :admin), flash: { error: result.value }
```
nemaniarjun: Can you tell me what will be the output of `result.value`? Perhaps we should have a more user-friendly message here.

piotrekpasciak: Error message depends on what ImportService returns:
- https://github.com/AIcrowd/AIcrowd/pull/1142/files#diff-5698dfbb2dc60b2bddc71e7a2068f349R8-R22

Issue #886, opened 2019-03-11
A participant can be an organizer in multiple challenges


spMohanty:
The participant model seems to simply store a single organizer id. I am assuming that means we cannot add a participant to multiple organizations.

mridulnagpal:
@spMohanty We can create a model to link between organizations and participants to deal with this situation. But we should confirm on what to do before making changes to the model. 
What do you recommend?

spMohanty:
@mridulnagpal  @nemaniarjun : Yes indeed. We cant have a separate field in the participant model link to a challenge, because a user can be an organizer in multiple organizations. Please go ahead with creating a model linking organizations and participants.

nemaniarjun:
Can we assign this to @bkDJ, I am not good with rails db stuff :sweat_smile:

PR #1145, opened 2020-02-18
Feature/participant has many organizer

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/458112901

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -45,7 +45,12 @@ class ChallengesController < ApplicationController
   end
 
   def new
-    @challenge = Challenge.new
+    @challenge = if params[:organizer_id]
+                   organizer = Organizer.find(params[:organizer_id])
+                   organizer.challenges.new
+                 else
+                   Challenge.new
+                 end
     authorize @challenge
   end
 
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -45,12 +45,8 @@ class ChallengesController < ApplicationController
   end
 
   def new
-    @challenge = if params[:organizer_id]
-                   organizer = Organizer.find(params[:organizer_id])
-                   organizer.challenges.new
-                 else
-                   Challenge.new
-                 end
+    @organizer = Organizer.find_by(params[:organizer_id])
+    @challenge = Challenge.new(organizer: @organizer)
     authorize @challenge
   end
 
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -45,7 +45,7 @@ class ChallengesController < ApplicationController
   end
 
   def new
-    @organizer = Organizer.find_by(params[:organizer_id])
+    @organizer = Organizer.find_by(id: params[:organizer_id])
     @challenge = Challenge.new(organizer: @organizer)
     authorize @challenge
   end
--- a/app/models/organizer.rb
+++ b/app/models/organizer.rb
@@ -6,8 +6,9 @@ class Organizer < ApplicationRecord
               use: [:slugged, :finders, :history]
   has_many :challenges,
            dependent: :destroy
+  has_many :participant_organizers
   has_many :participants,
-           dependent: :nullify
+           through: :participant_organizers
   has_many :clef_tasks,
            dependent: :destroy
   has_many :challenge_calls,
--- a/app/models/organizer.rb
+++ b/app/models/organizer.rb
@@ -6,7 +6,8 @@ class Organizer < ApplicationRecord
               use: [:slugged, :finders, :history]
   has_many :challenges,
            dependent: :destroy
-  has_many :participant_organizers
+  has_many :participant_organizers,
+           dependent: :destroy
   has_many :participants,
            through: :participant_organizers
   has_many :clef_tasks,
--- a/app/models/participant.rb
+++ b/app/models/participant.rb
@@ -23,7 +23,8 @@ class Participant < ApplicationRecord
          :omniauthable, omniauth_providers: %i[github oauth2_generic]
 
   default_scope { order('name ASC') }
-  belongs_to :organizer, optional: true
+  has_many :participant_organizers
+  has_many :organizers, through: :participant_organizers
   has_many :submissions, dependent: :nullify
   has_many :votes, dependent: :destroy
   has_many :blogs, dependent: :nullify
--- a/app/models/participant.rb
+++ b/app/models/participant.rb
@@ -23,7 +23,7 @@ class Participant < ApplicationRecord
          :omniauthable, omniauth_providers: %i[github oauth2_generic]
 
   default_scope { order('name ASC') }
-  has_many :participant_organizers
+  has_many :participant_organizers, dependent: :destroy
   has_many :organizers, through: :participant_organizers
   has_many :submissions, dependent: :nullify
   has_many :votes, dependent: :destroy
--- a/app/models/participant_organizer.rb
+++ b/app/models/participant_organizer.rb
@@ -1,0 +1,4 @@
+class ParticipantOrganizer < ApplicationRecord
+  belongs_to :participant
+  belongs_to :organizer
+end
--- a/app/models/participant_organizer.rb
+++ b/app/models/participant_organizer.rb
@@ -1,4 +1,5 @@
 class ParticipantOrganizer < ApplicationRecord
   belongs_to :participant
   belongs_to :organizer
+  validates :organizer_id, uniqueness: { scope: :participant_id }
 end
--- a/app/policies/challenge_policy.rb
+++ b/app/policies/challenge_policy.rb
@@ -157,9 +157,8 @@ class ChallengePolicy < ApplicationPolicy
       if participant&.admin?
         scope.all
       else
-        # need to fix for organizer
-        if participant&.organizer_id
-          scope.where("status_cd IN ('running','completed','starting_soon') OR organizer_id = ?", participant.organizer_id)
+        if participant&.organizers&.any?
+          scope.where("status_cd IN ('running','completed','starting_soon') OR organizer_id = ?", participant.organizer_ids)
         elsif participant
           scope.where(participant_sql(email: participant.email))
         else
--- a/app/policies/challenge_policy.rb
+++ b/app/policies/challenge_policy.rb
@@ -8,7 +8,7 @@ class ChallengePolicy < ApplicationPolicy
   end
 
   def edit?
-    participant && (participant.admin? || @record.organizer_id == participant.organizer_id)
+    participant && (participant.admin? || participant.organizers.ids.include?(@record.organizer_id))
   end
 
   def reorder?
@@ -24,7 +24,7 @@ class ChallengePolicy < ApplicationPolicy
   end
 
   def new?
-    participant && (participant.admin? || participant.organizer_id.present?)
+    participant && (participant.admin? || participant.organizers.present?)
   end
 
   def create?
@@ -81,12 +81,12 @@ class ChallengePolicy < ApplicationPolicy
     @record.challenge_rounds.present? &&
       @record.show_leaderboard == true ||
       (participant &&
-          (participant.admin? || @record.organizer_id == participant.organizer_id))
+          (participant.admin? ||  participant.organizers.ids.include?(@record.organizer_id)))
   end
 
   def submissions_allowed?
     return false unless @record.online_submissions
-    return true if participant && (participant.admin? || @record.organizer_id == participant.organizer_id)
+    return true if participant && (participant.admin? ||  participant.organizers.ids.include?(@record.organizer_id))
 
     if @record.running? || (@record.completed? && @record.post_challenge_submissions?)
       if @record.clef_challenge.present?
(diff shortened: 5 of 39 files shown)


db/migrate/20200214101258_add_participant_organizer.rb
```diff
@@ -0,0 +1,10 @@
+class AddParticipantOrganizer < ActiveRecord::Migration[5.2]
+  def change
+    # move existing data of participant's organizer_id into middle table (ParticipantOrganizer)
+    participants = Participant.where.not(organizer_id: nil)
```
piotrekpasciak: This kind of operations are usually done by rake task or data_migrations. Can you move it to rake task?

db/schema.rb
```diff
@@ -10,11 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_02_10_143323) do
+ActiveRecord::Schema.define(version: 2020_02_14_101258) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "pg_stat_statements"
+  enable_extension "pg_trgm"
```
piotrekpasciak: Are you sure this line should be present? It looks like something from your local setup. Local database changes shouldn't be committed into schema.rb, because when other person runs migration, they will get change that removes that extension from schema.rb.

spec/factories/participant_factory.rb
```diff
@@ -10,7 +10,7 @@
     address                              { FFaker::Address.street_address }
     city                                 { FFaker::Address.city }
     country_cd                           { FFaker::Address.country_code }
-    organizer                            { nil }
+    # organizer                            { nil }
```
piotrekpasciak: It's a bad habit to leave commented code. Just remove it.

spec/factories/participant_organizers.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :participant_organizer do
+    participant { nil }
```
piotrekpasciak: Default data should be more like:
association(:participant)
association(:organizer)

spec/models/participant_organizer_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ParticipantOrganizer, type: :model do
```
piotrekpasciak: Empty `*_spec.rb` should not be commited to repository.

piotrekpasciak:
There are still some failings specs. Please take a look.

app/policies/participant_policy.rb
```diff
@@ -44,7 +44,8 @@ def show_pending_invitations?
   end
 
   def clef_access?
-    participant && (participant.admin? || participant.id == @record.id || participant.organizer && participant.organizer.clef_organizer == true)
+    # can't use first org
+    participant && (participant.admin? || participant.id == @record.id || participant.organizers && participant.organizers.first.clef_organizer == true)
```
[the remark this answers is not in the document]
nemaniarjun: I don't think we need clef_access as a permission for the user. 
Any user that has edit access to a challenge should have clef access to that challenge.

app/models/participant_organizer.rb
```diff
@@ -0,0 +1,4 @@
+class ParticipantOrganizer < ApplicationRecord
```
nemaniarjun: Are you sure we need this model? 
Organizer (should be called organization really)  has a direct relationship with Participant. we can use has_and_belongs_to for the relation

app/controllers/challenges_controller.rb
```diff
@@ -44,7 +44,12 @@ def show
   end
 
   def new
-    @challenge = Challenge.new
+    @challenge = if params[:organizer_id]
```
nemaniarjun: This could be written more clearly, also what happens if I pass a organizer_id which does not exist or has been deleted?

nemaniarjun: Suggestion:
```ruby
def new
    @organizer = Organizer.find_by(params[:organizer_id]) # Returns nil if not found
    @challenge = Challenge.new(organizer: @organizer) # organizer: nil is the default == same if record not found
    authorize @challenge
end
```

spec/controllers/clef_tasks_controller_spec.rb
```diff
@@ -3,15 +3,16 @@
...
-  let!(:clef_tasks)        { create_list(:clef_task, 3, organizer: organizer) }
+  let!(:clef_tasks)           { create_list(:clef_task, 3, organizer: organizer) }
 
-  let(:organizer)          { create :organizer }
-  let(:organizer_admin)    { create :participant, organizer: organizer }
-  let(:valid_attributes)   { FactoryBot.attributes_for(:clef_task) }
-  let(:invalid_attributes) { FactoryBot.attributes_for(:clef_task, :invalid) }
+  let(:organizer)             { create :organizer }
+  let(:organizer_admin)       { create :participant }
+  let(:participant_organizer) { create(:participant_organizer, participant: organizer_admin, organizer: organizer)}
```
nemaniarjun: I still dont see the need for this model.

app/models/participant_organizer.rb
piotrekpasciak: `has_and_belongs_to` is an old syntax and should not be used anymore:
https://github.com/rubocop-hq/rails-style-guide#user-content-has-many-through

Having this model is very useful, when browsing database through rails console. Additionally we can implement needed validations here. I'm strongly in favor of that every database table should have corresponding model.

db/migrate/20200214084730_create_participant_organizers.rb
```diff
@@ -0,0 +1,10 @@
+class CreateParticipantOrganizers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :participant_organizers do |t|
```
piotrekpasciak: You should also add `uniqueness` index and uniqueness validation in model (for participant_id scoped to organizer_id). This way we'll prevent duplicated records.

app/models/participant_organizer.rb
nemaniarjun: That does make sense, but I really don't like the naming scheme. We should change Organizer -> Organization and ParticipantOrganizer to Organizer. Doesn't have to be in this PR.

piotrekpasciak: @nemaniarjun I'm against changing the name of table and model the way you proposed. Join table should be named `participants_organizers` and model. ParticipantsOrganizer. Your proposition is. against Rails way.

alkeshrails: Agree with @piotrekpasciak, If we change model name than other developer confuse in it.

app/policies/challenge_policy.rb
```diff
@@ -145,6 +145,7 @@ def resolve
       if participant&.admin?
         scope.all
       else
+        # need to fix for organizer
```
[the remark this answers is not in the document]
alkeshrails: Done!

app/policies/organizer_policy.rb
```diff
@@ -55,6 +55,7 @@ def resolve
       if participant&.admin?
         scope.all
       else
+        # need to fix for organizer
```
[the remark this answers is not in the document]
alkeshrails: Done!

app/policies/participant_policy.rb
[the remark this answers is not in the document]
alkeshrails: I also check code base and this method is not calling from anywhere, I removed it.

app/models/participant_organizer.rb
alkeshrails: Agree with @piotrekpasciak, If we change model name than other developer confuse in it.

app/policies/submission_policy.rb
```diff
@@ -83,13 +83,13 @@ def resolve
...
       else
-        if participant&.organizer_id
+        if participant&.organizers.present?
           sql = %[
             #{participant_sql(participant)}
             OR challenge_id IN
               (SELECT c.id
                 FROM challenges c
-                WHERE c.organizer_id = #{participant.organizer_id})
+                WHERE c.organizer_id = #{participant.organizers.first.id})
```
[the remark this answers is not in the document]
alkeshrails: done!

db/migrate/20200214084730_create_participant_organizers.rb
alkeshrails: Done!

db/migrate/20200214101258_add_participant_organizer.rb
alkeshrails: Okay, moving.

db/schema.rb
alkeshrails: Done!

app/policies/participant_policy.rb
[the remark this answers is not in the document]
piotrekpasciak: Just remove that method if it's not used anywhere.

app/policies/challenge_policy.rb
```diff
@@ -149,8 +149,8 @@ def resolve
       if participant&.admin?
         scope.all
       else
-        if participant&.organizer_id
-          scope.where("status_cd IN ('running','completed','starting_soon') OR organizer_id = ?", participant.organizer_id)
+        if participant&.organizers&.any?
+          scope.where("status_cd IN ('running','completed','starting_soon') OR organizer_id = ?", participant.organizer_ids)
```
piotrekpasciak: Shouldn't that be changed to `organizer_id IN []`? Will it work correctly in current version?

db/migrate/20200214084730_create_participant_organizers.rb
piotrekpasciak: Uniqueness index is still missing.

spec/features/challenges/challenge_datasets_spec.rb
```diff
@@ -8,7 +8,8 @@
   let!(:challenge_participant)           { create :challenge_participant, challenge: challenge, participant: participant }
   let!(:admin)                           { create :participant, :admin }
   let!(:challenge_admin_participant)     { create :challenge_participant, challenge: challenge, participant: admin }
-  let!(:organizer)                       { create :participant, organizer: challenge.organizer }
+  let!(:organizer)                       { create :participant }
+  let(:participant_organizer)            { create(:participant_organizer, organizer: challenge.organizer, participant: organizer)}
```
piotrekpasciak: I think you should be able to use `let!` here instead to not change log_in method below.

alkeshrails:
> most of it looks great.
> I just don't see the point of having participant_organizers as a separate model.

We are using participant_organizers as middle table between participant and organizers. If we want to change Organizer to Organization than we need t change in existing code base.

app/controllers/challenges_controller.rb

spec/controllers/clef_tasks_controller_spec.rb
alkeshrails: Here I use this for create relation between Participant and Organizer.

db/migrate/20200214084730_create_participant_organizers.rb
alkeshrails: Added!

app/policies/challenge_policy.rb
alkeshrails: Done!

spec/features/challenges/challenge_datasets_spec.rb
alkeshrails: Done!

piotrekpasciak:
@alkeshrails You still need to handle at least following places:
https://github.com/AIcrowd/AIcrowd/blob/master/app/controllers/organizers_controller.rb#L24
https://github.com/AIcrowd/AIcrowd/blob/master/app/helpers/organizers_helper.rb#L5
https://github.com/AIcrowd/AIcrowd/blob/master/app/helpers/s3_files_helper.rb#L38
https://github.com/AIcrowd/AIcrowd/blob/master/app/policies/member_policy.rb#L3
https://github.com/AIcrowd/AIcrowd/blob/master/app/policies/member_policy.rb#L43
https://github.com/AIcrowd/AIcrowd/blob/master/app/policies/participant_challenge_policy.rb#L31
https://github.com/AIcrowd/AIcrowd/blob/master/app/policies/task_dataset_file_policy.rb#L7
https://github.com/AIcrowd/AIcrowd/blob/master/app/views/challenges/tabs/_details_form.html.erb#L75
https://github.com/AIcrowd/AIcrowd/blob/master/app/views/challenges/tabs/_details_form.html.erb#L76

!IMPORTANT! - probably my list is not complete - so you have to browse the codebase by yourself.

app/models/participant.rb
```diff
@@ -23,7 +23,8 @@ class Participant < ApplicationRecord
          :omniauthable, omniauth_providers: %i[github oauth2_generic]
 
   default_scope { order('name ASC') }
-  belongs_to :organizer, optional: true
+  has_many :participant_organizers
```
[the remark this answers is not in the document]
alkeshrails: Done!

app/models/organizer.rb
```diff
@@ -6,8 +6,9 @@ class Organizer < ApplicationRecord
               use: [:slugged, :finders, :history]
   has_many :challenges,
            dependent: :destroy
+  has_many :participant_organizers
```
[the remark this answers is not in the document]
alkeshrails: Done!

nemaniarjun:
Closes https://github.com/AIcrowd/AIcrowd/issues/886

PR #1150, opened 2020-02-18
Update order in challenges listing

piotrekpasciak:


nemaniarjun:
I am not sure if this is what was needed. Sometimes we have challenges that might need to be featured on the landing page even if they are closed.

@skbly7 can you please confirm. 

We can also always use a different scope / ordering for the landing page.

piotrekpasciak:
@skbly7 @spMohanty 
Should challenges order fix affect only challenges listing?
https://www.aicrowd.com/challenges

piotrekpasciak:
Actually not needed.

PR #1152, opened 2020-02-19
Changes for reputation and rating system

nikhilrayaprolu:
Here are the schema changes and the active job that calls https://github.com/AIcrowd/trueskill-service for rating calculation.

Please review in detail as this is my first PR in rails code, need to learn some of the best practices possible around.

The active job is activated in a cron job every night at 2:10 am to calculate both permanent and temporary scores of the users into the database.

Also, how do you decide which indexes to be created in the database?

Environment variable to be added:
RATING_API_URL: "https://mighty-spire-92101.herokuapp.com"
To reset ratings:
```
ChallengeRound.all().update_all(calculated_permanent: false)
UserRating.delete_all()
Participant.all().update_all(rating: nil, variation: nil)
Participant.all().update_all(temporary_rating: nil, temporary_variation: nil)
```


app/models/user_rating.rb
```diff
@@ -0,0 +1,4 @@
+class UserRating < ApplicationRecord
+  belongs_to :participant
+  belongs_to :challenge_round
```
alkeshrails: Please add appropriate association on the other end also (On participant and challenge_round model).

app/services/trueskill_service.rb
```diff
@@ -0,0 +1,25 @@
...
+  def call(ranks, teams_mu, teams_sigma, teams_participant_ids)
+    @body = {
+        :ranks => ranks,
+        :teams_mu => teams_mu,
+        :teams_sigma => teams_sigma,
+        :teams_participant_ids => teams_participant_ids
+    }.to_json
+    @headers = { 'Content-Type' => 'application/json' }
+    # Add http://localhost:8888/calculate for local testing of the api
+    @result = HTTParty.post('/calculate', :body => @body, :headers => @headers)
```
piotrekpasciak: You're sending request to external service. Consider handling network errors.

Gemfile
```diff
@@ -11,6 +11,8 @@ gem 'rails-assets-jQuery-File-Upload'
 ruby '2.5.5'
 gem 'rails', '5.2.2.1'
 
+gem 'activerecord-import'
```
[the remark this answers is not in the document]
nikhilrayaprolu: Yes, we are using them to activerecord-import for bulk import - optimises bulk import

and rufus scheduler is used because of these issues:
https://stackoverflow.com/questions/50625601/sidekiqcronjob-throw-me-uninitialized-constant-rufusschedulercronline-er
https://github.com/ondrejbartas/sidekiq-cron/issues/199

db/migrate/20200218064815_create_user_ratings.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserRatings < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_ratings do |t|
+      t.references :participant, foreign_key: true
+      t.float :rating
```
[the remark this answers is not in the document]
nikhilrayaprolu: this basically keeps the user rating history required to show the time series graph - https://sketch.cloud/s/vEeEQ/a/OYx4oR
whereas the participant table is used to retrieve the latest value.

db/migrate/20200217201127_add_columns_to_participant.rb
```diff
@@ -0,0 +1,6 @@
+class AddColumnsToParticipant < ActiveRecord::Migration[5.2]
+  def change
+    add_column :participants, :rating, :float
```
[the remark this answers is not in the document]
nikhilrayaprolu: Yes, I took the decision by looking at https://stackoverflow.com/questions/8514167/float-vs-decimal-in-activerecord
In fact, we don't do any calculations of float numbers in ruby, all the calculations are handled by the python service.

app/jobs/rating_calculate_job.rb
```diff
@@ -0,0 +1,17 @@
+class RatingCalculateJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    sorted_challenge_rounds = ChallengeRound.where("calculated_permanent=FALSE OR calculated_permanent is NULL").order(end_dttm: :desc)
```
skbly7: It should be `desc` or `asc`? I am assuming this code is generating rating after end of every challengeRound, in that case the oldest round should be rated first and so on.

app/jobs/rating_calculate_job.rb
```diff
@@ -0,0 +1,17 @@
...
+    Participant.update_all(temporary_rating: nil)
+    Participant.update_all(temporary_variation: nil)
+    for round in sorted_challenge_rounds
+      if round.end_dttm.present? && round.end_dttm < Time.current
+        UserRatingService.new(round, false).call
```
skbly7: _For the future, if we aren't handling it right now_
How should we handle the ChallengeRound ending extension when rating is already calculated, followed by rating of rounds which ended after it?

app/jobs/rating_calculate_job.rb
skbly7: _For the future, if we aren't handling it right now_
What about hidden challenges?

app/queries/leaderboard_participants_query.rb
```diff
@@ -0,0 +1,22 @@
+class LeaderboardParticipantsQuery
+  attr_reader :relation
+
+  def initialize(relation = Leaderboard.all)
+    @relation = relation
+  end
+
+  def participants_with_rating(challenge_round)
```
skbly7: What is happening here? The query is too complex to understand, can we do it in Rails native way? If not, let's write a comment at least explaining query intent.

app/services/trueskill_service.rb
```diff
@@ -0,0 +1,33 @@
...
+    body = {
+        :ranks => ranks,
+        :teams_mu => teams_mu,
+        :teams_sigma => teams_sigma,
+        :teams_participant_ids => teams_participant_ids
+    }.to_json
+    headers = { 'Content-Type' => 'application/json' }
+    # Add http://localhost:8888/calculate for local testing of the api
+    begin
+      result = HTTParty.post('http://localhost:8888/calculate', :body => body, :headers => headers)
```
skbly7: Get `GRADER_SERVICE_ENDPOINT` or so from the environment variable, and clean localhost references.

db/schema.rb
```diff
@@ -305,8 +305,9 @@
     t.text "leaderboard_note"
     t.integer "failed_submissions", default: 0
     t.integer "parallel_submissions", default: 0, null: false
-    t.string "score_title", default: "Score", null: false
-    t.string "score_secondary_title", default: "Secondary Score", null: false
+    t.string "score_title", default: "", null: false
```
skbly7: Why the default values have been changed in this schema.rb for `Score` and `Secondary Score`?  Please undo it.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    participant_ids.each_with_index do |participant_id, i|
+      if @temporary
+        userratings << UserRating.new(participant_id: participant_id, temporary_rating: new_participant_ratings[i]['rating'].to_f, temporary_variation: new_participant_variations[i]['variation'].to_f, challenge_round: @round)
+      else
+        userratings << UserRating.new(participant_id: participant_id, rating: new_participant_ratings[i]['rating'].to_f, variation: new_participant_variations[i]['variation'].to_f, challenge_round: @round)
+      end
+    end
+    UserRating.import userratings
+  end
+  def call()
```
skbly7: Is this function atomic? If not, let's make it. There can be lot of cases when calculation breaks mid-way due to whatsoever reason, leaving DB in an inconsistent state.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    leaderboard_rating_stats = LeaderboardParticipantsQuery.new.participants_with_rating @round.id
```
skbly7: If we already know `@temporary` it will be good to fetch only `temperory_mu` & `temporary_sigma` (and vice versa when not temperory), when making SQL query in LeaderboardParticipantsQuery.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+    return final_team_temporary_mu, final_team_temporary_sigma
+  end
+  def filter_leaderboard_stats(leaderboard_rating_stats)
+    ranks, teams_mu, teams_sigma, teams_participant_ids = [], [], [], []
+    teams_temporary_mu, teams_temporary_sigma = [], []
+    for team in leaderboard_rating_stats
+      begin
+        teams_participant_ids << team['participant_id'].split(',')
+      rescue NoMethodError
```
piotrekpasciak: You should avoid rescue blocks like that.  They can catch way more errors then intended. If you run into risk that `team['participant_id']` is nil, use guard class e.g `next if team['participant_id'].blank?`

https://devblast.com/b/what-are-guard-clauses

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,87 @@
...
+        userratings << UserRating.new(participant_id: participant_id, temporary_rating: new_participant_ratings[i]['rating'].to_f, temporary_variation: new_participant_variations[i]['variation'].to_f, challenge_round: @round)
+      else
+        userratings << UserRating.new(participant_id: participant_id, rating: new_participant_ratings[i]['rating'].to_f, variation: new_participant_variations[i]['variation'].to_f, challenge_round: @round)
+      end
+    end
+    UserRating.import userratings
+  end
+  def call()
+    leaderboard_rating_stats = LeaderboardParticipantsQuery.new.participants_with_rating @round.id
+    ranks, teams_mu, teams_sigma, teams_participant_ids, teams_temporary_mu, teams_temporary_sigma = self.filter_leaderboard_stats leaderboard_rating_stats
```
piotrekpasciak: Are u sure `self.` is necessary? I think you can call methods from this class without `.self`

app/jobs/rating_calculate_job.rb
nikhilrayaprolu: Can you elaborate, I didn't understand.

app/jobs/rating_calculate_job.rb
nikhilrayaprolu: you are right, changing the code

nikhilrayaprolu: How do you want to handle them? Currently, they are considered as a normal challenge.

app/queries/leaderboard_participants_query.rb
nikhilrayaprolu: Sure, I will comment it, this is the maximum railsy way, I could think of for this complex query.

app/services/trueskill_service.rb
nikhilrayaprolu: Yep, this is a mistake in the recent commit, editing out

db/schema.rb
nikhilrayaprolu: that's something that happened while merging the master, changing it

app/services/user_rating_service.rb
nikhilrayaprolu: modified the query and the code accordingly

app/models/challenge_round.rb
```diff
@@ -4,7 +4,9 @@ class ChallengeRound < ApplicationRecord
   belongs_to :challenge, inverse_of: :challenge_rounds
   has_many :submissions, dependent: :restrict_with_error
   has_many :leaderboards, dependent: :destroy
-
+  has_many :user_ratings
+  after_initialize :defaults,
```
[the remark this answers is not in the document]
nikhilrayaprolu: this is one of mistakes that occured while merging the changes from master

app/services/trueskill_service.rb
```diff
@@ -0,0 +1,33 @@
...
+    rescue HTTParty::Error => e
+      Rails.logger.error e.message
+    rescue StandardError => e
```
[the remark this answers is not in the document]
nikhilrayaprolu: sometimes whenever the external API service is down, StandardError gets raised, so in those scenarios, we have to know that the API service is down and the only way to know about it is to log whenever the API service is not available.

Let me know if you have any other suggestions

app/services/user_rating_service.rb
nikhilrayaprolu: you are right @piotrekpasciak

app/services/user_rating_service.rb
nikhilrayaprolu: changing the code accordings, thanks @piotrekpasciak

app/services/user_rating_service.rb
nikhilrayaprolu: made it atomic using ActiveRecord Transactions

app/jobs/rating_calculate_job.rb
skbly7: What will happen when the challenges deadline is extended? 

1. Challenge ended at 10:00pm.
2. Rating calculated in platform at 2:30am next day.
3. Organiser/AIcrowd decided to extend the challenge for 1 more week.
4. \<??\>

app/jobs/rating_calculate_job.rb
skbly7: I am not sure, that is why it is an open question for you w.r.t. rating feature design. 😉  

We can consider them as normal challenges, for now, I personally feel hidden ones to unrated or something.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,66 @@
...
+    @round = round
+    @temporary = temporary
+  end
+
+  def call()
+    ActiveRecord::Base.transaction do
+      leaderboard_rating_stats = LeaderboardParticipantsQuery.new.participants_with_rating @round.id, @temporary
+      ranks, teams_mu, teams_sigma, teams_participant_ids = filter_leaderboard_stats leaderboard_rating_stats
+      if ranks.length > 1
+        new_team_ratings, new_team_variations = TrueskillService.new.call ranks, teams_mu, teams_sigma, teams_participant_ids
```
skbly7: What if `TrueskillService.new.call` returns nil due to Error in its API calls? Looks like this code block will continue to run & end up marking everything as done instead of raising error/queuing calculation for the current challenge round again.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,66 @@
...
+            new_participant_ratings << {"rating" => new_team_ratings[i][j] }
+            new_participant_variations << {"variation" => new_team_variations[i][j]}
+          end
+        end
+      end
+      return participant_ids, new_participant_ratings, new_participant_variations
+    end
+
+    def update_database_columns(participant_ids, new_participant_ratings, new_participant_variations)
+      Participant.update participant_ids, new_participant_ratings
```
skbly7: You would need to use `update!` instead of `update` so errors can be raised causing the transaction to break, otherwise, it wouldn't break on errors.

> This is why we use the .update! instead of just .update. Since a normal update doesn’t raise errors and just return “false”, it would not trigger a rollback in the transaction.

https://medium.com/@kristenrogers.kr75/rails-transactions-the-complete-guide-7b5c00c604fc

cc: @piotrekpasciak for throwing in his experience with transaction here.

app/jobs/rating_calculate_job.rb
nikhilrayaprolu: This requires a bit of brainstorming in future if the below assumption is proved to be wrong,
why?
Let x be a challenge. As the final rating has a recurrent dependency on all previous challenges, if we have calculated 10 other challenge scores after the calculation of challenge x, we have to rollback scores to the state when all these 11 challenges (10 + x) are not calculated and recalculate the rating affected by x and all these 10 challenges.

Current code considers the final rating calculated in the previous day and then uses the same rating to calculate the rating affected by the challenge again (So the challenge is considered twice). This assumption is partly true because Trueskill assumes that as the users participates in 'n' no.of.challenges, their current rating starts to reflect their "true rating", So the change in current ratings is negligible after 'n' challenges. Therefore, even if the challenge scores are considered twice there won't be much change in the ratings after their current rating has reached the "true rating".

app/jobs/rating_calculate_job.rb
nikhilrayaprolu: What is the business use case of `hidden challenges`? depending on that we can decide

app/services/user_rating_service.rb
nikhilrayaprolu: fixed this issue

app/services/user_rating_service.rb
nikhilrayaprolu: @piotrekpasciak @skbly7   I get `NoMethodError` for `update!`
image

app/services/trueskill_service.rb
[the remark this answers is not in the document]
piotrekpasciak: I've never encountered a situation when unavailable API would raise `StandardError`. Maybe you mean that connection error would be raised. Consider following list of network errors:
 [
    Timeout::Error, Errno::ECONNRESET, Errno::ECONNREFUSED,
    Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError, Net::ProtocolError
  ]

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,70 @@
+class UserRatingService
+  def initialize(round, temporary)
+    @round = round
+    @temporary = temporary
+  end
+
+  def call()
+    ActiveRecord::Base.transaction do
```
piotrekpasciak: This transaction is very risky. It may generate a lot of deadlocks because u're doing requests to external API within transaction and they take long time. Placing transaction here requires much more thought here and must be done wisely. @skbly7 How would you approach this problem?

app/services/user_rating_service.rb
piotrekpasciak: @skbly7 In comment below I wrote this transaction is very dangerous as it may generate a lot of deadlocks. It needs to be placed mutch smarter.

app/services/trueskill_service.rb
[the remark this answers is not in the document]
nikhilrayaprolu: I came up with code using these two links,
https://stackoverflow.com/questions/7909596/how-can-i-handle-errors-with-httparty/7909823
https://stackoverflow.com/questions/39944374/how-to-handle-httparty-errors-rails

Do you want me to Implement all these list of Network errors individually or just put Standard Error as all these inherit from StandardError and can be caught with one line?

app/services/user_rating_service.rb
nikhilrayaprolu: resolved by moving all database updates into one function and making it atomic

app/services/trueskill_service.rb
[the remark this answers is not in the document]
piotrekpasciak: Those 2 links are prime example you cannot trust stackoverflow and informations from internet.

I definetly want you to cach list of Network errors as catching `StandardError` catches far more errors then Network Errors. If you. leave StandardError we won't be able notice other. issues with code like Nil or Argument exceptions.

app/services/user_rating_service.rb
nikhilrayaprolu: updated the transaction to only database updates. Now there won't any such troubles from deadlocks

piotrekpasciak: That makes situation mutch better, but experience tells me we may still encounter deadlocks. We'll monitor for Rollbar and take action when we encounter them.

app/services/trueskill_service.rb
```diff
@@ -0,0 +1,31 @@
+class TrueskillService
+  include HTTParty
+  debug_output $stdout
+  base_uri ENV["TRUESKILL"]
```
skbly7: The renaming task needs to be done before this can be merged!

nikhilrayaprolu:
image

image

Here are the new changes for UI

@skbly7 @alkeshrails @piotrekpasciak @jsnrynlds

app/controllers/participants_controller.rb
```diff
@@ -21,6 +21,14 @@ def show
                                Discourse::FetchUserPostsService.new(participant: @participant).call
                              end
     @discourse_posts = @discourse_posts_fetch.value
+
+    @categories = @participant.challenges.collect{
```
nikhilrayaprolu: the collect inside collect would make 100 SQL requests for receiving category names (as they are in a different table and are not provided by default in the challenge model) if the user has participated in 100 challenges.

Other way round would be to use an inner join using joins()  for the participant with challenges and then categories and just a group by gives us the counts in a dictionary.  - So that we could get the counts and joins handled in one SQL query.

As a participant with 100 challenges is not foreseen in the near future (Not sure, if Educational Challenges are considered), I don't think it is a trouble, so we could merge the pull request.

skbly7: Make sense, updated the code with efficient version of the query.

app/queries/leaderboard_participants_query.rb
```diff
@@ -0,0 +1,30 @@
+class LeaderboardParticipantsQuery
+  attr_reader :relation
+
+  def initialize(relation = Leaderboard.all)
+    @relation = relation
+  end
+
+  def participants_with_rating(challenge_round, temporary)
```
piotrekpasciak: We should use `call` as only public method apart of `initialize` for QueryObject classes to keep code consistent.

app/queries/participant_rating_ranks_query.rb
```diff
@@ -0,0 +1,41 @@
+class ParticipantRatingRanksQuery
+  attr_reader :relation
+
+  def initialize(relation = Participant.all)
+    @relation = relation
+  end
+
+  def participants_with_ranks()
```
piotrekpasciak: QueryObject should contain single method named `call`. Please refactor this class into multiple. Consider using namespaces.

app/queries/participant_rating_ranks_query.rb
```diff
@@ -0,0 +1,41 @@
...
+    # this query joins leaderboard with team participants if the submission is a team based submission, or else with the individual
+    # participants, then it calls for rating and variation fields, whenever rating from a team based user is not available, it searches
+    # for a rating from a participant based submission and if even a participant based submission is not available it then uses 0 as default
+    # Also, for temporary fields we take permanent rating as default value whenever a temporary rating is not available.
+    @relation.where('rating is not null and ranking > 0').select('slug, name, rating, variation, created_at, ranking, ranking_change').reorder('ranking asc')
+  end
+
+  def participants_ranks_update()
+    ranks = @relation.where('rating is not null').select('id, rank() over (order by rating desc) as current_ranking, ranking as previous_ranking, ranking_change').reorder('ranking asc')
+    participant_rankings = ranks.map {|participant| {'ranking' => participant.current_ranking}}
```
piotrekpasciak: Avoid rocket syntax for hashes:
https://github.com/rubocop-hq/ruby-style-guide#hash-literals

app/services/rating_api_service.rb
```diff
@@ -0,0 +1,26 @@
...
+  def call(challenge_rounds)
+    body = {
+        rounds: challenge_rounds
+    }.to_json
+    headers = { 'Content-Type' => 'application/json' }
+    begin
+      result = self.class.post('/startroundcalculations/', body: body, headers: headers)
+    rescue HTTParty::Error => e
+      Rails.logger.error e.message
+    rescue StandardError => e
```
piotrekpasciak: Please specify errors you want to resuce. Catching all errors like `rescue StandardError => e` is just asking for trouble.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,66 @@
+class UserRatingService
```
piotrekpasciak: Service Object class should implement only single public method called `call`. apart of constructor.  Other methods should be private. Please write basic test coverage for this service, at least for positive path.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,66 @@
...
+      userratings = []
+      participant_ids.each_with_index do |participant_id, i|
+        if @temporary
+          userratings << UserRating.new(participant_id: participant_id, temporary_rating: new_participant_ratings[i]['temporary_rating'].to_f, temporary_variation: new_participant_variations[i]['temporary_variation'].to_f, challenge_round: @round)
+        else
+          userratings << UserRating.new(participant_id: participant_id, rating: new_participant_ratings[i]['rating'].to_f, variation: new_participant_variations[i]['variation'].to_f, challenge_round: @round)
+        end
+      end
+      UserRating.import userratings
+      if not @temporary
```
piotrekpasciak: Please don't use `not` operator as it has low precedence. Use `unless @temporary ` instead.

app/services/user_rating_service.rb
```diff
@@ -0,0 +1,66 @@
...
+        @round.update_attribute :calculated_permanent, true
```
piotrekpasciak: Why do you use `update_attribute` here? Are you sure you want to skip validations and callbacks? It's not recommended to do so. 

Please use `update!` if possible.

app/views/blogs/_blog_post.html.erb
```diff
@@ -7,8 +7,7 @@
     </div>
 
     <div class="card-footer">
-      <%= link_to image_tag(blog.participant.image_url, class: 'avatar'), participant_path(blog.participant)%>
-      <span>By  <%= link_to blog.participant.name, participant_path(blog.participant) %></span>
+      <%= render :partial => 'participants/user_link', locals: { participant: blog.participant } %>
```
piotrekpasciak: Use new hash syntax.

app/views/blogs/show.html.erb
```diff
@@ -15,10 +15,8 @@
     <div class="row">
       <div class="col-12 d-flex align-items-center flex-wrap">
         <div class="author">
-          <%= link_to image_tag(@blog.participant.image_url, class: 'avatar'), participant_path(@blog.participant)%>
-          <span>
-            By <%= link_to @blog.participant.name, participant_path(@blog.participant) %>
-          </span>
+          <span class="mr-2">By</span>
+          <%= render :partial => 'participants/user_link', locals: { participant: @blog.participant } %>
```
piotrekpasciak: New hash syntax.

app/views/challenges/_clef_participants.html.erb
```diff
@@ -4,10 +4,7 @@
   <li class="list-group-item">
     <% if participant.present? %>
       <div>
-        <%= link_to participant_path(participant) do %>
-          <%= image_tag participant.image_url, class: 'avatar' %>
-          <span class="text-link"><%= participant.name %></span>
-        <% end %>
+        <%= render :partial => 'participants/user_link', locals: { participant: participant } %>
```
piotrekpasciak: Hash syntax.

app/views/challenges/teams/_member_list.html.erb
```diff
@@ -16,12 +16,9 @@
     <ul class="list-group list-group-flush-px-0 list-group-teams">
       <% members.each do |tp| %>
         <li class="list-group-item">
-          <%= link_to tp.participant do %>
-            <%= image_tag tp.participant.image_url, class: "avatar" %>
-            <span class="text-link"><%= tp.participant.name %></span>
-            <% if tp.role_organizer? %>
-              <span class="badge badge-success"><%= t('.members.organizer_badge') %></span>
-            <% end %>
+          <%= render :partial => 'participants/user_link', locals: { participant: tp.participant } %>
```
piotrekpasciak: New hash syntax for whole file.

app/views/landing_page/_blog_post.html.erb
```diff
@@ -7,8 +7,7 @@
 	  </div>
 
 	  <div class="card-footer">
-	      <%= link_to image_tag(post.participant.image_url, class: 'avatar'), participant_path(post.participant)%>
-	      <span>By <%= link_to post.participant.name, participant_path(post.participant) %></span>
+	    <%= render :partial => 'participants/user_link', locals: { participant: post.participant } %>
```
piotrekpasciak: Hash syntax

app/views/landing_page/index.html.erb
```diff
@@ -92,30 +92,28 @@
...
         <div class="card">
           <ul class="list-group list-group-flush list-group-latest-contributors">
             <% @participants.each do |participant| %>
-              <!-- user -->
               <li class="list-group-item">
-                <%= link_to participant_path(participant) do %>
-                  <%= image_tag participant.image_url, class: "avatar" %>
-                  <span class="text-link"><%= participant.name %></span>
-                <% end %>
+                <%= render :partial => 'participants/user_link', locals: { participant: participant } %>
```
piotrekpasciak: Hash syntax

app/views/layouts/_header.html.erb
```diff
@@ -32,6 +32,9 @@
 					<li class="nav-item">
 						<%= link_to 'Challenges', challenges_path, {:class => "nav-link"} %>
 					</li>
+          <li class="nav-item">
+            <%= link_to 'Leaderboard', leaderboard_path, {:class => "nav-link"} %>
```
piotrekpasciak: New hash syntax for whole file.

app/views/leaderboards/_leaderboards.html.erb
```diff
@@ -15,10 +15,7 @@
         <%= image_tag default_image_url, class: 'avatar' %>
         <span class="text-link">Unknown User</span>
       <% elsif participants.length == 1 %>
-        <%= link_to participant_path(participants[0]), title: participants[0].name, data: {toggle: :tooltip} do %>
-          <%= image_tag participants[0].image_url, class: 'avatar' %>
-          <span class="text-link"><%= participants[0].name %></span>
-        <% end %>
+        <%= render :partial => 'participants/user_link', locals: { participant: participants[0] } %>
```
piotrekpasciak: Hash syntax

app/views/members/_members.html.erb
```diff
@@ -13,10 +13,7 @@
           <% @members.each do |member| %>
             <!-- team member -->
             <li class="list-group-item">
-              <%= link_to participant_path(member) do %>
-                <%= image_tag member.image_url, class: "avatar" %>
-                <span class="text-link"><%= member.name %></span>
-              <% end %>
+              <%= render :partial => 'participants/user_link', locals: { participant: member } %>
```
piotrekpasciak: Hash syntax

app/views/participant_challenges/_participant_challenge_rows.erb
```diff
@@ -1,9 +1,6 @@
 <% @participant_challenges.each do |challenge_participant| %>
   <% participant = challenge_participant.participant %>
   <li class="list-group-item">
-    <%= link_to participant_path(participant) do %>
-      <%= image_tag participant.image_url, class: 'avatar' %>
-      <span class="text-link"><%= participant.name %></span>
-    <% end %>
+    <%= render :partial => 'participants/user_link', locals: { participant: participant } %>
```
piotrekpasciak: Hash syntax

app/views/participants/_challenge_list.html.erb
```diff
@@ -0,0 +1,78 @@
...
+              <% end %>
+            </div>
+            <% if challenge.tagline %>
+              <p class="card-text"><%= challenge.tagline %></p>
+            <% end %>
+            <%= render partial: 'shared/challenges/prizes_breakdown', locals: { challenge: challenge } %>
+        </div>
+    </div>
+    <div class="col-md-5 mt-n2">
+      <% submissions = challenge.submissions.where(participant_id:participant.id).reorder('created_at DESC').limit(3) %>
```
piotrekpasciak: Database queries should not be placed in view file. Please move it to controller.

app/views/participants/_challenge_list.html.erb
```diff
@@ -0,0 +1,78 @@
...
+      <header class="section-header">
+        <h4>Latest submissions</h4>
+        <% if submissions.any? %>
+          <a class="btn btn-secondary btn-sm" style="float:right" href="<%= challenge_submissions_path(challenge) %>?q%5Bparticipant_name_equals%5D=<%= participant.name %>">See All</a>
```
piotrekpasciak: Using HTML style attribute is very bad practice. Try `pull-right` bootstrap css class.

app/views/participants/_masthead.html.erb
```diff
@@ -4,12 +4,13 @@
...
         <div class="d-flex">
-          <%= image_tag participant.image_url, class: 'avatar' %>
+          <%= participant_avatar participant %>
           <div class="masthead-content">
             <div class="d-flex">
-              <%= image_tag participant.image_url, class: "avatar-sm" %>
+              <%= participant_avatar participant, "avatar-sm" %>
               <h1>
                 <%= participant.name %>
+                <%= render :partial => 'participants/show_reputation', locals: { participant: participant } %>
```
piotrekpasciak: New hash syntax for whole file.

app/views/participants/_show_reputation.erb
```diff
@@ -0,0 +1,17 @@
+<span class="reputation-info" style="<%= style ||= '' %>">
+<% if participant.admin %>
+  <span class="reputation-score" aria-label="Admin" 
+      data-balloon-pos="up">
+    ###
+  </span>
+<% else %>
+  <span class="reputation-score" aria-label="<%= participant.rating.to_i %> ± <%= participant.variation.to_i %>" 
+      data-balloon-pos="up">
+    <% if defined?(full) and full %>
```
piotrekpasciak: Please use `&&` instead of `and`.

app/views/participants/_tooltip_profile.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="tooltip-inner-top">
+<%= participant_avatar participant %>
+    <div>
+        <strong><%= participant.name %></strong>
+        <br>
+        <div class="reputation-info">
+            <%= render :partial => 'participants/show_reputation', locals: { participant: participant, style: 'top: 0px;', full: true } %>
```
piotrekpasciak: New hash syntax for whole file.

app/views/participants/_user_link.html.erb
```diff
@@ -0,0 +1,17 @@
+<% 
+    uniq_id = rand(1000)
```
piotrekpasciak: Using `rand` is no very unique. There is significant risk of collision. Why do you need this logic in first place?

If you absolutely need this unique value, please use something reliable like `SecureRandom.uuid`

app/views/participants/_user_link.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    random_class = 'tooltip-user-wrapper-' + uniq_id.to_s + '-' + participant.id.to_s
+%>
+<div class="tooltip-user-wrapper <%= random_class %>">
+    <div style="display: table-row;" data-container=".<%= random_class %>" data-html="true" data-toggle="tooltip" 
+        title='<%= render partial: 'participants/tooltip_profile', locals: { participant: participant } %>'>
+        <a href="<%= participant_path(participant.name) %>">
+            <% if not defined?(avatar) or (defined?(avatar) and avatar) %>
```
piotrekpasciak: Please don't use `not`, `or`, `and`. Use !, ||, && instead.

app/views/rating_leaderboard/_rating_leaderboard.html.erb
```diff
@@ -0,0 +1,30 @@
+<%
+  participants.each do |participant|
+  if @current_participant and @current_participant.name == participant.name
```
piotrekpasciak: Use && instead of `and`.

app/views/rating_leaderboard/index.html.erb
```diff
@@ -0,0 +1,54 @@
...
+                  <div>
+                    <span id="more-anchor"></span>
+                    <%= render partial: 'shared/load_more', locals: { collection: @top_participants } %>
+                  </div>
+                </td>
+              </tr>
+            </tbody>
+            <% if @self_standing %>
+            <thead>
+              <tr><th colspan="6" style="text-align: center;">Your Standings</th></tr>
```
piotrekpasciak: Please don't use `style` attribute. You could use `text-center` bootstrap class here.

app/views/shared/utility/_flag.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if country_cd.present? %>
+    <img src="/assets/flags/<%= country_cd %>.svg" style="width: 24px" />
```
piotrekpasciak: Please, move this style to CSS class.

app/views/submissions/_submissions.html.erb
```diff
@@ -9,7 +9,7 @@
     <td><strong><%= submission.id %></strong></td>
     <td>
       <% if participant.present? %>
-        <%= link_to participant.name, participant_path(participant) %>
+        <%= render :partial => 'participants/user_link', locals: { participant: participant, avatar: false } %>
```
piotrekpasciak: New hash syntax.

db/migrate/20200217153133_add_calculated_permanent_to_challenge_round.rb
```diff
@@ -0,0 +1,5 @@
+class AddCalculatedPermanentToChallengeRound < ActiveRecord::Migration[5.2]
+  def change
+    add_column :challenge_rounds, :calcualated_permanent, :boolean
```
piotrekpasciak: Please add `null: false` constraint. Code is mutch more reliable when you don't have to handle. NULL everywhere.

db/schema.rb
```diff
@@ -208,6 +208,31 @@
     t.index ["creator_id"], name: "index_blazer_queries_on_creator_id"
   end
 
+  create_table "blogit_comments", force: :cascade do |t|
```
piotrekpasciak: Are you sue that `blogit_commits`. should be part of this PR?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_09_134625) do
+ActiveRecord::Schema.define(version: 2020_03_14_060809) do
```
piotrekpasciak: Please tidy up schema.rb to include changes only from migrations in this PR.

Gemfile
```diff
@@ -11,12 +11,15 @@ gem 'rails-assets-jQuery-File-Upload'
...
+gem 'rufus-scheduler', '~> 3.4.0'
 gem 'ahoy_matey'
 gem 'blazer', github: "aicrowd/blazer"
 gem 'active_model_serializers'
 gem 'activeadmin', '1.4.3'
 gem 'authtrail'
 gem 'aws-sdk', '~> 2.2'
+gem 'bootstrap', '~> 4.2.1'
```
piotrekpasciak: We should use Yarn to install javascript and CSS libraries, so we can update libraries whenever new version shows up in NPM, not when gem maintainers update. the gem.

app/helpers/rating_leaderboard_helper.rb
```diff
@@ -0,0 +1,37 @@
+module RatingLeaderboardHelper
+  def tier_calculate(ranking)
+    percentile = (1 - ((ranking - 1).to_f/@count))*100
```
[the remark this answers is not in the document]
nikhilrayaprolu: @count is an instance variable in the rating_leaderboard_controller, as the helper gets included in the controller it's available even in the helper.

app/helpers/rating_leaderboard_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def rating_leaderboard_ranking_change(participant)
+    if !(participant.ranking_change.nil? || participant.ranking_change == 0)
+      if participant.ranking_change > 0
+        image_tag(
+            "icon-change-up.svg",
+            data:  { toggle: 'tooltip' },
+            title: "+#{participant.ranking_change} change")
+      else
```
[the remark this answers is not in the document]
nikhilrayaprolu: It could be reached when the `ranking_change` is 0, and it's also working in the leaderboard view.

app/helpers/rating_leaderboard_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+    when 60..80
+      tier = 4
+    when 0..60
+      tier = 5
+    end
+  end
+
+  def rating_leaderboard_ranking_change(participant)
+    if !(participant.ranking_change.nil? || participant.ranking_change == 0)
+      if participant.ranking_change > 0
```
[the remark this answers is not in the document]
nikhilrayaprolu: Sure, but we don't need .positive as `ranking_change` could also be negative.

Gemfile
piotrekpasciak: We've discussed that we leave Bootstrap in Gemfile for now. We'll move it to Yarn, after we properly migrate to Webpack.

app/models/participant.rb
```diff
@@ -160,6 +166,25 @@ def image_url
                 end
   end
 
+  def rating_tier_class
```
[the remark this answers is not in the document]
nikhilrayaprolu: @skbly7

app/helpers/rating_leaderboard_helper.rb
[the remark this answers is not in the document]
piotrekpasciak: Please, pass this variable as parameter to this method e.g `tier_calculate(ranking, count)`

app/helpers/rating_leaderboard_helper.rb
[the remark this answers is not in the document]
piotrekpasciak: You're right, didn't take into account situation when raking_change would be negative.

app/helpers/rating_leaderboard_helper.rb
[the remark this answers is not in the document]
piotrekpasciak: You're right, didn't take into account situation when raking_change would be negative.

app/services/user_rating_service.rb
nikhilrayaprolu: for service pattern, I hadn't come across such a rule that it should have only one single public function I had come across patterns where there are multiple public functions in almost all of the opensource rails applications. Also, it doesn't make sense to have a separate class for each function when what they do is similar,
some other files where service pattern has multiple functions include:
https://github.com/discourse/discourse/blob/master/app/services/notification_emailer.rb
https://github.com/discourse/discourse/blob/master/app/services/post_alerter.rb

Here in this same repo we have
https://github.com/AIcrowd/AIcrowd/blob/master/app/services/s3_service.rb

app/services/user_rating_service.rb
```diff
@@ -59,7 +59,7 @@ def update_database_columns(participant_ids, new_participant_ratings, new_partic
       end
       UserRating.import userratings
       unless @temporary
-        @round.update_attribute :calculated_permanent, true
+        @round.update(calculated_permanent: true)
```
piotrekpasciak: It's strongly recommended to use `update!` instead of `update` to not swallow errors silently.

app/models/participant.rb
```diff
@@ -111,6 +111,10 @@ def reserved_userhandle
     errors.add(:name, 'is reserved for CrowdAI users.  Please log in via CrowdAI to claim this user handle.') if (provider != 'crowdai') && ReservedUserhandle.where(name: name.downcase).exists?
   end
 
+  def rated_users
```
piotrekpasciak: That should be Ruby on Rails scope:
https://www.rubyguides.com/2019/10/scopes-in-ruby-on-rails/

PR #1156, opened 2020-02-21
Create invitaion to update form

alkeshrails:
Invite participant view should be fixed, also have comma seperated in text (along with csv_
URL : https://aicrowd.monday.com/boards/363093650/pulses/457955933


app/controllers/challenges_controller.rb
```diff
@@ -148,6 +149,12 @@ def set_s3_direct_post
                             acl:                   'private')
   end
 
+  def create_invitations
+    params[:challenge][:invitation_email].split(',').each do |email|
+      @challenge.invitations.create(email: email.strip)
```
piotrekpasciak: You should use methods with `!` whenever it makes sense to follow policy of failing fast. Here it would be `create!`.

app/views/challenges/fields/_invitations_table_fields.html.erb
```diff
@@ -3,13 +3,12 @@
...
-            f,
-            :invitations,
-            html_options = {
-              partial: 'challenges/fields/invitation_fields',
-              class: 'btn btn-secondary btn-sm' } %>
+  <div>
+    <div class="form-group">
+      <%= f.label :invitation_email, 'Email addresses' %>
+      <%= f.text_field :invitation_email, placeholder: 'Email address', class: 'form-control', 'aria-describedby': 'emailHelp' %>
+      <small id="emailHelp" class="form-text text-muted">Separate Email by comma (,)</small><br/>
```
piotrekpasciak: `emails` *

app/controllers/challenges_controller.rb
```diff
@@ -65,6 +65,7 @@ def edit; end
 
   def update
     if @challenge.update(challenge_params)
+      create_invitations if params[:challenge][:invitation_email]
```
piotrekpasciak: `params[:challenge][:invitation_email].present?`, because empty string is truthy in Ruby.

app/controllers/challenges_controller.rb

app/views/challenges/tabs/_private_challenge_form.html.erb
```diff
@@ -3,15 +3,28 @@
...
       <%= form_tag import_challenge_invitations_path(@challenge), multipart: true do %>
         <%= file_field_tag :file %>
         <%= submit_tag "Invite Participants from CSV", class: 'btn' %>
       <% end %>
     </div>
   <% end %>
+  <%= render partial: 'challenges/fields/submit_buttons', locals: { f: f, next_step: 'submissions' } %>
 </div>
+<script type="text/javascript">
+  $(document).ready(function(){
```
skbly7: Instead of document.ready it should be called when the form is re-rendered. Otherwise, the button isn't working currently when we try to invite multiple times.

Usecase:
1. Open invite participant and click Invite button
2. Add email and save.
3. After update, go back to participant tab & click invite button
4. (Currently) Button isn't clickable, (Desired) As happened after 1st

Check out how we are handling JS in other places for challenge edit page, @piotrekpasciak or @nemaniarjun can help.

alkeshrails: In my case, request is Html. So I can't use this. I change logic here. Now I call ajax on `invite` button click. Have a look.

alkeshrails: I change it, because I don't thing call a ajax for just showing input email field, is good idea. I enable this functionality to Jquery. Please have a look.

app/models/challenge.rb
```diff
@@ -85,6 +85,8 @@ class Challenge < ApplicationRecord
     :challenge_round_summaries
   ].freeze
 
+  attr_accessor :invitation_email
```
piotrekpasciak: Is this attt_accessor necessary?

spec/features/challenges/private_challenge_spec.rb
```diff
@@ -1,5 +1,22 @@
...
+      log_in(participant)
+      visit "/challenges/#{challenge.id}/edit?step=private-challenge"
+    end
+
+    it "Invite participant by email" do
+      click_button '+ Invite participant'
+      fill_in 'challenge[invitation_email]', with: 'participant1@user.com, participant2@user.com'
+
+      click_button 'Update challenge'
+      expect(page).to have_text "Challenge updated."
```
piotrekpasciak: Could you also add following assertion? `expect(page).to have_content "participant1@user.com"

If there will be issue with it, I'll take a look.

alkeshrails: Done!

Issue #1088, opened 2020-01-20
Add leaderboard & submissions export functionality for organisers


skbly7:
There are multiple requests from organisers who require the export for submissions & leaderboards over time. Currently we are doing it manually and passing on the exported files. 

This should be full fledged feature which we should have instead on rails app side.

skbly7:
Completed by #1157 and #1159.

PR #1157, opened 2020-02-21
Export leaderboards to CSV file

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/432235176

Github issue:
https://github.com/AIcrowd/AIcrowd/issues/1088

### DIFF
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -153,6 +153,15 @@ Paloma.controller('Challenges', {
           }, 1000);
         });
 
+        $('#leaderboard-export-rounds-select').on('change', function(event) {
+          const leaderboardExportChallengeRoundId = event.target.value;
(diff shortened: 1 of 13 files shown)


app/policies/leaderboard_policy.rb
```diff
@@ -18,6 +18,10 @@ def submission_detail?
       )
   end
 
+  def export?
+    true
```
nemaniarjun: Should regular (non-admin non-organizer) participants allowed to export leaderboards?

piotrekpasciak: @nemaniarjun i've updated authorization. rules and specs to follow the same rules as challenge export to json.

skbly7:
This PR has already been merged, but I was able to test it just now.

The exported leaderboard have duplicated/repeated content due to some error. This need to be fixed in follow up PR.

Example (from staging):

image

Exported CSV (4 times repeated):

image

PR #1158, opened 2020-02-24
Fix/improve remove participant view

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/470753007
Improve Remove Participant view

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -1,11 +1,7 @@
 class ChallengesController < ApplicationController
   before_action :authenticate_participant!, except: [:show, :index]
   before_action :terminate_challenge, only: [:show, :index]
-<<<<<<< HEAD
-  before_action :set_challenge, only: [:show, :edit, :update, :destroy, :clef_task, :remove_image, :export, :import]
-=======
-  before_action :set_challenge, only: [:show, :edit, :update, :destroy, :clef_task, :remove_image, :export, :remove_invited]
->>>>>>> Remove invited participant
+  before_action :set_challenge, only: [:show, :edit, :update, :destroy, :clef_task, :remove_image, :export, :import, :remove_invited]
   before_action :set_vote, only: [:show, :clef_task]
   before_action :set_follow, only: [:show, :clef_task]
   after_action :verify_authorized, except: [:index, :show]
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -131,7 +131,8 @@ class ChallengesController < ApplicationController
   end
 
   def remove_invited
-    @challenge.invitations.find_by(id: params[:invited_id]).destroy
+    challenge_invitation = @challenge.invitations.find(params[:invited_id])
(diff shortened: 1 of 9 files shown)


app/controllers/challenges_controller.rb
```diff
@@ -115,6 +115,10 @@ def export
               filename: "#{@challenge.challenge.to_s.parameterize.underscore}_export.json"
   end
 
+  def remove_invited
+    @challenge.invitations.find_by(id: params[:invited_id]).destroy
```
piotrekpasciak: We should avoid possibility of `Nil exception` here. Consider following code:
`challenge_invitation = @challenge.invitations.find(params[:invited_id])`
`challenge_invitation.destroy!`

app/views/challenges/remove_invited.js.erb
```diff
@@ -0,0 +1,3 @@
+$('.remove_invited').bind('ajax:success', function(){
```
piotrekpasciak: We should use `id` instead of `class` to make code more reliable.

app/views/challenges/fields/_invitations_table_fields.html.erb
```diff
@@ -23,14 +23,14 @@
...
-            <td><%= invitation.object.created_at %></td>
-            <td><%= link_to_remove_association "remove", invitation %></td>
-          </tr>
-        <% end %>
+        <tr>
+          <td><%= invitation.object.email %></td>
+          <td><%= invitation.object.invitee_name %></td>
+          <td><%= participant_invitation(email: invitation.object.email) %></td>
+          <td><%= invitation.object.created_at %></td>
+          <td><%= link_to "remove", remove_invited_challenge_path(@challenge, invited_id: invitation.object.id), class: 'remove_invited', remote: true %></td>
```
piotrekpasciak: Please use longer, more descriptive class name to avoid colission e.g `challenge-invitation-removte-invited`

alkeshrails: Done!

app/controllers/challenges_controller.rb

spec/features/challenges/private_challenge_spec.rb
```diff
@@ -19,4 +20,16 @@
...
+
+  context "Logged in user should remove invited participant" do
+    before do
+      log_in(participant)
+      visit "/challenges/#{challenge.id}/edit?step=private-challenge"
+    end
+    it "Remove invited participant to remove link" do
+      click_link 'remove'
+
+      expect(page).to have_no_content 'invited@user.com'
```
piotrekpasciak: This test assertions are not very reliable. You should also check http status and current_path. Consider this example:

        expect(page).to have_http_status 200
        expect(page).to have_current_path challenge_leaderboards_path(challenge)

alkeshrails: done!

alkeshrails: done!

PR #1159, opened 2020-02-24
Add submissions export to CSV functionality

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/432235176

Github issue:
https://github.com/AIcrowd/AIcrowd/issues/1088


app/services/submissions/csv_export_service.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    private
+
+    attr_reader :submissions, :meta_keys
+
+    def csv_headers
+      [
+        'Submission ID',
+        'Participant',
+        'Score',
```
skbly7: Is it possible to make `Score` & `Secondary Score` dynamic? And use `score_title` and `score_secondary_title` field for names.

app/services/submissions/csv_export_service.rb
```diff
@@ -0,0 +1,70 @@
...
+        'Secondary Score',
+        'Description',
+        *meta_headers,
+        'Post Challenge',
+        'Created At',
+        'Updated At',
+        'Grading Message',
+        'Grading Status',
+        'Challenge Round'
+      ]
```
skbly7: Need to contain:

```
type (team/participant)
team_name (“-” when not a team)
participants_list
```

Similar to leaderboard export feature.

app/views/challenges/tabs/_admin_form.html.erb
```diff
@@ -15,6 +15,10 @@
     <%= link_to 'Export Leaderboard to CSV', export_challenge_leaderboards_url(challenge, leaderboard_export_challenge_round_id: challenge_rounds.first&.id.to_i), id: 'leaderboard-export-link', class: 'btn btn-primary' %>
   </div>
 
+  <div class="form-group">
+    <%= link_to 'Export Submissions to CSV', export_challenge_submissions_path(challenge), class: 'btn btn-primary' %>
```
skbly7: Unable to differentiate with above form right now. Better to add atleast label for now:

```
<label>Submissions Export</label>
```

app/views/challenges/tabs/_admin_form.html.erb
```diff
@@ -15,6 +15,10 @@
     <%= link_to 'Export Leaderboard to CSV', export_challenge_leaderboards_url(challenge, leaderboard_export_challenge_round_id: challenge_rounds.first&.id.to_i), id: 'leaderboard-export-link', class: 'btn btn-primary' %>
```
skbly7: Oops! Admin form is only available to admin and not organisers. I wasn't aware of the same while suggesting to use "Admin" tab. We will need to move the latest export features to a new tab "Misc".

app/services/submissions/csv_export_service.rb
piotrekpasciak: We export submissions for multiple challenge rounds, so do we want to display first challenge_round name or some other way?

skbly7: You are right, those names can vary. 
Will it be straight forward to add a "round" dropdown for submissions as well (like we do for leaderboard)? That can solve our problem.

piotrekpasciak: Yes, I can do that.

nemaniarjun:
Do we need the Admin tab to be hidden from organizers?

skbly7:
I am not sure about the intended usage of the admin panel here, but the current behavior is hidden.

But I think having one "admin"-only tab anyways make sense (even if currently we don't have such settings which only we want to control), so better to move new features into a new tab (Misc) irrespective of the above decision.

app/services/submissions/csv_export_service.rb
piotrekpasciak: @skbly7 Are you. sure about that? Submissions belongs only to single participant:

app/views/challenges/tabs/_admin_form.html.erb
piotrekpasciak: I've added heading instead as label didn't look well.

app/services/submissions/csv_export_service.rb
skbly7: Yes, lets add team information. We anyway create leaderboard model based on submissions only, you can check that sidekiq job to know how it is being deduced.

piotrekpasciak: @skbly7 Could you send me a link? I'm having trouble finding it.

PR #1160, opened 2020-02-24
Fix link in submission media partial

piotrekpasciak:


app/views/shared/challenges/_media.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_type = media_content_type(mediable) %>
 
 <% if ['video','image','youtube'].include?(content_type) && size == :thumb %>
-  <%= link_to challenge_submission_path(mediable.challenge_id, mediable.id) do %>
+  <%= link_to challenge_submission_path(mediable.challenge_id, submission_id) do %>
     <%= media_asset(mediable, content_type, size) %>
```
nemaniarjun: Instead of sending it as a local, mediable.submission_id should also work right?

piotrekpasciak: I don't think so, this partial is used in two different places. In one place `mediable` is `Leaderboard` instance and in other it is `Submission` instance.

PR #1168, opened 2020-02-25
Sanitise users html inputs before rendering

piotrekpasciak:
Monday: 
https://aicrowd.monday.com/boards/363093650/pulses/458027248

Read up:
https://makandracards.com/makandra/2579-everything-you-know-about-html_safe-is-wrong


app/helpers/application_helper.rb
```diff
@@ -39,7 +39,7 @@ def footer_class
   def themed_button(opts = {})
     raise ArgumentError if opts.key?(:link) && opts.key?(:modal)
 
-    title = opts[:title].presence || '&nbsp;'.html_safe
+    title = opts[:title].presence
```
nemaniarjun: there should be fallback to '&nbsp;' ?

app/helpers/devise_helper.rb
```diff
@@ -15,17 +15,16 @@ def devise_error_messages!
...
-      #{messages}
-      <button type="button" class="close" data-dismiss="alert" aria-label="Close">
-        <span aria-hidden="true">&times;</span>
-      </button>
-    </div>
-  </div>
-    HTML
 
-    html.html_safe
+    html =  "<div class='alert alert-warning alert-dismissible fade show' role='alert'>"
```
nemaniarjun: no html_safe for the first line?

app/helpers/application_helper.rb
piotrekpasciak: I guess I can change `presence` method to `to_s`  to achieve it.

app/helpers/devise_helper.rb
piotrekpasciak: This method looks different now.

app/views/dataset_files/_toc_form.html.erb
```diff
@@ -1,27 +0,0 @@
-<br/><br/><br/>
```
skbly7: I don't find refactoring in every PR attractive approach. 

I understand it is tempting to get rid of files that aren't in use/dead code, but that needs to be separate PRs. This causes reviewing the actual/important changes meant via current PR harder.

spec/helpers/sanitization_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
skbly7: Nice! 💯

app/views/dataset_files/_toc_form.html.erb
piotrekpasciak: I've moved removals to separate PR, but this one is blocked until we merge the new one:
https://github.com/AIcrowd/AIcrowd/pull/1169

piotrekpasciak:
@skbly7 Code removal is no longer part of this PR. Please take a look once again.

PR #1171, opened 2020-02-26
Add datatables to challenge invitations table

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/458107017

Github issue:
https://github.com/AIcrowd/AIcrowd/issues/64

### DIFF
--- a/app/javascript/stimulus_js/controllers/simple_datatables_controller.js
+++ b/app/javascript/stimulus_js/controllers/simple_datatables_controller.js
@@ -1,0 +1,9 @@
+import { Controller } from 'stimulus';
+import $ from 'jquery';
+
+export default class extends Controller {
+  connect() {
+    const tableSelector = this.element.getAttribute('data-table-id');
+    $(`#${tableSelector}`).dataTable();
+  }
+}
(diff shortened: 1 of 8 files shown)


app/javascript/stimulus_js/controllers/simple_datatables_controller.js
```diff
@@ -0,0 +1,12 @@
+require('datatables');
+
+import { Controller } from 'stimulus';
+import $ from 'jquery';
```
nemaniarjun: I am worried about multiple inclusions of jquery in the final generated pack.
Jquery is already loaded via a CDN to the application layout, do we need to import it here?

nemaniarjun: http://techs.studyhorror.com/rails-jquery-loaded-twice-application-js-i-168

piotrekpasciak: Yes, we need to import it here. Otherwise won't be able to use `$` in code.

PR #1181, opened 2020-02-28
Multiple organizers support for one challenge

piotrekpasciak:
Rake Task:
`RAILS_ENV=production bundle exec rake challenges:migrate_challenge_organizer_to_challenges_organizers_table`

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -56,6 +56,7 @@ class ChallengesController < ApplicationController
   def create
     @challenge                = Challenge.new(challenge_params)
     @challenge.clef_challenge = true if @challenge.organizers.any? { |organizer| organizer.clef_organizer? }
+    @challenge.organizers     = current_participant.organizers if current_participant&.admin? == false
 
     authorize @challenge
 
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -6,7 +6,7 @@ class ChallengesController < ApplicationController
   before_action :set_follow, only: [:show, :clef_task]
   after_action :verify_authorized, except: [:index, :show]
   before_action :set_s3_direct_post, only: [:edit, :update]
-  before_action :set_organizer, only: [:edit, :update]
+  before_action :set_organizers, only: [:edit, :update]
   before_action :set_organizers_for_select, only: [:new, :create]
   before_action :set_challenge_rounds, only: [:edit, :update]
   before_action :set_category, only: [:new, :create , :edit]
@@ -47,13 +47,15 @@ class ChallengesController < ApplicationController
 
   def new
     @organizer = Organizer.find_by(id: params[:organizer_id])
-    @challenge = Challenge.new(organizer: @organizer)
+    @challenge = Challenge.new
+    @challenge.organizers << @organizer if @organizer.present?
+
     authorize @challenge
   end
 
   def create
     @challenge                = Challenge.new(challenge_params)
-    @challenge.clef_challenge = true if @challenge.organizer&.clef_organizer
+    @challenge.clef_challenge = true if @challenge.organizers.any? { |organizer| organizer.clef_organizer? }
 
     authorize @challenge
 
@@ -122,7 +124,7 @@ class ChallengesController < ApplicationController
   end
 
   def import
-    result = Challenges::ImportService.new(import_file: params[:import_file], organizer: @challenge.organizer).call
+    result = Challenges::ImportService.new(import_file: params[:import_file], organizers: @challenge.organizers).call
 
     if result.success?
       redirect_to edit_challenge_path(@challenge, step: :admin), notice: 'New Challenge Imported'
@@ -139,7 +141,7 @@ class ChallengesController < ApplicationController
   private
 
   def set_challenge
-    @challenge = Challenge.includes(:organizer).friendly.find(params[:id])
+    @challenge = Challenge.includes(:organizers).friendly.find(params[:id])
     @challenge = @challenge.versions[params[:version].to_i].reify if params[:version]
     authorize @challenge
   end
@@ -152,8 +154,8 @@ class ChallengesController < ApplicationController
     @follow = @challenge.follows.where(participant_id: current_participant.id).first if current_participant.present?
   end
 
-  def set_organizer
-    @organizer = @challenge&.organizer
+  def set_organizers
+    @organizers = @challenge&.organizers
   end
 
   def set_organizers_for_select
@@ -232,11 +234,13 @@ class ChallengesController < ApplicationController
       :dynamic_content_tab,
       :dynamic_content,
       :dynamic_content_url,
-      :organizer_id,
       image_attributes: [
         :id,
         :image,
         :_destroy
+      ],
+      challenges_organizers_attributes: [
+        :organizer_id
       ],
       challenge_partners_attributes: [
         :id,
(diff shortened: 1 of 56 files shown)


app/policies/challenge_policy.rb
```diff
@@ -8,7 +8,7 @@ def show?
   end
 
   def edit?
-    participant && (participant.admin? || @record.organizer_id == participant.organizer_id)
+    participant && (participant.admin? || @record.organizers.ids.include?(participant.organizer_id))
```
alkeshrails: I also add similar type condition, it will do conflicts.
Check here : https://github.com/AIcrowd/AIcrowd/pull/1145/files#diff-9ef788e882df06ec909aeae72fa104c0R11

app/policies/submission_policy.rb
```diff
@@ -9,7 +9,7 @@ def show?
 
     @record.challenge.submissions_page.present? && (
       (participant && (participant.admin? ||
-        @record.challenge.organizer_id == participant.organizer_id)) ||
+        @record.challenge.organizers.ids.include?(participant.organizer_id))) ||
```
alkeshrails: same as above.

app/policies/challenge_policy.rb
piotrekpasciak: Yeah, but someone needs to merge the PR first, so we can resolve the conflicts.

nemaniarjun:
Closes https://github.com/AIcrowd/AIcrowd/issues/8

app/controllers/challenges_controller.rb
```diff
@@ -51,7 +51,7 @@ def new
 
   def create
     @challenge                = Challenge.new(challenge_params)
-    @challenge.clef_challenge = true if @challenge.organizer&.clef_organizer
+    @challenge.clef_challenge = true if @challenge.organizers.any? { |organizer| organizer.clef_organizer? }
```
nemaniarjun: We should consider moving the clef_challenge part on challenges instead of participants or challenge_organizers, if you feel this same, we should create a new issue for that.

app/models/submission.rb
```diff
@@ -20,20 +20,9 @@ class Submission < ApplicationRecord
 
   validates :challenge_id, presence: true
   validates :grading_status, presence: true
-  validate :clef_validations
 
   delegate :name, :email, to: :participant, allow_nil: true
 
-  def clef_validations
-    return true
-    return true unless challenge.organizer.clef?
```
nemaniarjun: @ieggel is it a requirement for each clef submission to have a method_description of length > 5?
We seem to have disabled the code at the top with `return true`
let us know if we need this validation!

app/policies/challenge_policy.rb
```diff
@@ -80,13 +80,12 @@ def has_accepted_participation_terms?
   def show_leaderboard?
     @record.challenge_rounds.present? &&
       @record.show_leaderboard == true ||
-      (participant &&
-          (participant.admin? || @record.organizer_id == participant.organizer_id))
+      (participant && (participant.admin? || @record.organizers.ids.include?(participant.organizer_id)))
```
nemaniarjun: the last part can be replaced by ` show_leaderboard == true || edit?`

app/policies/challenge_policy.rb
```diff
@@ -80,13 +80,12 @@ def has_accepted_participation_terms?
...
   end
 
   def submissions_allowed?
     return false unless @record.online_submissions
-    return true if participant && (participant.admin? || @record.organizer_id == participant.organizer_id)
+    return true if participant && (participant.admin? || @record.organizers.ids.include?(participant.organizer_id))
```
nemaniarjun: same as above `return true if edit?`

app/policies/clef_task_policy.rb
```diff
@@ -1 +1,5 @@
-class ClefTaskPolicy < ChallengePolicy; end
+class ClefTaskPolicy < ChallengePolicy
```
nemaniarjun: This already inherits from challenge policy, why is this function defined again ? 
Are we not planning on supporting multiple organizers for clef challenges? Why ?

app/policies/submission_policy.rb
nemaniarjun: again we can reuse edit? policy

app/policies/ongoing_leaderboard_policy.rb
```diff
@@ -45,13 +45,20 @@ def resolve
...
       else
         if participant&.organizer_id
-          sql = %[
-            #{participant_sql(participant)}
-            OR challenge_id IN
-              (SELECT c.id
-                FROM challenges c
-                WHERE c.organizer_id = #{participant.organizer_id})
-          ]
+          challenges_ids = Challenge.left_joins(:organizers)
```
nemaniarjun: This is mostly the same as for other policy files, can we DRY this part out somehow? maybe as a separate query somewhere?

lib/tasks/challenges.rake
```diff
@@ -18,6 +18,16 @@ namespace :challenges do
     end
   end
 
+  desc 'Migrate challenge.organizer to challenges_organizers table'
+  task migrate_challenge_organizer_to_challenges_organizers_table: :environment do
+    Challenge.where('organizer_id IS NOT NULL').find_each do |challenge|
```
nemaniarjun:
```suggestion
    Challenge.where.not(organizer: nil).find_each do |challenge|
```

lib/tasks/challenges.rake
```diff
@@ -18,6 +18,16 @@ namespace :challenges do
...
+      ChallengesOrganizer.create!(challenge_id: challenge.id, organizer_id: challenge.organizer_id)
+      puts "Challenge##{challenge.id} finished migration of organizer"
+    end
+
+    puts "Challegne organizer migration comppleted."
```
nemaniarjun: typo in the word 'challenge' and in 'completed'

app/models/submission.rb
ieggel: HI @nemaniarjun ,
Yes it would be better to have a method_description with length > 5. 
Thank you!

piotrekpasciak: For some reason this validation was disabled. Now we have to make all records in database pass this validation before we can enable it.

app/policies/clef_task_policy.rb
piotrekpasciak: This change is needed here because `@record` in  `ChallengePolicy` is a `Challenge` instance and `@record` in `ClefTaskPolicy` is a `ClefTask` instance.

app/models/submission.rb
piotrekpasciak: I think we should do that in separate PR, as it may require rake task to readd this validation.

app/views/challenges/tabs/_details_form.html.erb
```diff
@@ -69,11 +69,19 @@
...
     <span class="label-text">Organizer</span>
-    <% if @challenge.new_record? && current_participant&.admin? %>
-      <%= f.select :organizer_id, @organizers_for_select, { include_blank: true } %>
+    <% if @challenge.new_record? %>
+      <% if current_participant&.admin? %>
+        <%= f.fields_for :challenges_organizers, [@challenge.challenges_organizers.new] do |challenges_organizer_form| %>
+          <%= challenges_organizer_form.select :organizer_id, @organizers_for_select, {}, class: 'form-control' %>
+        <% end %>
+      <% else %>
+        <% organizer_id = @challenge.organizers.first&.id || current_participant.organizer_id %>
```
alkeshrails: Do we just update challenge for its first organizer?

piotrekpasciak: You're right this. doesn't make sense. I've moved assigning the organizers to controller.

PR #1190, opened 2020-02-29
Tabular overview page bugfixes and moved code to controller

skbly7:
* Mobile rendering issue is now fixed
* Page resizes trigger rendering


app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -182,10 +182,80 @@ Paloma.controller('Challenges', {
           }
         });
       };
+
+      let table_of_content_as_tabs = function () {
+        if ($('#description-wrapper .md-content').length > 0) {
+          let force_scroll = false;
+          let content = $('#description-wrapper .md-content')[0].innerHTML;
+          $('<style>.force-active { color: #DE4B46 !important; font-weight: 700;}</style>').appendTo('body');
+          let first_link_in_content = $('#table-of-contents a').first().attr('href').split('#')[1];
```
piotrekpasciak: `first_link_in_content` seems to not be mutated anywhere. This kind of variables should be declared with `const` keyword instead of `let`.

Following variables also seem to not be mutated:
- first_link_content
- notification_data

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -182,10 +182,80 @@ Paloma.controller('Challenges', {
           }
         });
       };
+
+      let table_of_content_as_tabs = function () {
+        if ($('#description-wrapper .md-content').length > 0) {
+          let force_scroll = false;
+          let content = $('#description-wrapper .md-content')[0].innerHTML;
+          $('<style>.force-active { color: #DE4B46 !important; font-weight: 700;}</style>').appendTo('body');
```
piotrekpasciak: Appending additional css to body is very unusual approach. Standard approach would to define css class and add it to selected elements.

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -182,10 +182,80 @@ Paloma.controller('Challenges', {
...
+          let first_link_in_content = $('#table-of-contents a').first().attr('href').split('#')[1];
+          let first_link_content = $(content).filter('h2#'+first_link_in_content)[0].outerHTML;
```
piotrekpasciak: const

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -182,10 +182,80 @@ Paloma.controller('Challenges', {
...
+          let notification_data = content.split(first_link_content)[0];
+          let reset_content_to_everything_if_mobile = function() {
```
piotrekpasciak: More standard approach is to move function declaration outside of business logic e.g top / bottom of file. 

With webpack we would be able to move this declaration to helper file that would be imported. But we can't do that now, unfortunately.

nemaniarjun:
@skbly7 would you like me to take this up ? I can fix it and merge it into master so that we have it parallel to prod?

app/javascript/stimulus_js/controllers/challenge_overview_controller.js
```diff
@@ -0,0 +1,100 @@
...
+                .each( (i,x) => content += x.outerHTML );
+        } else {
+            $(start).nextUntil(' h2 ')
+                .addBack() // Add back the 'start' element to the array
+                .each( (i,x) => content += x.outerHTML );
+        }
+
+        return content;
+    }
+in
```
piotrekpasciak: Should `in` be here?

app/javascript/stimulus_js/controllers/challenge_overview_controller.js
```diff
@@ -0,0 +1,100 @@
...
+
+    createTOC(){
+        $.each(this.headings, (index, heading) => {
+            // Add mt-0 to headings
+            $(heading).addClass('mt-0');
+            // Add id
+            $(heading).attr('id', `heading-${index}`);
+
+            // Create new list item in the TOC
+            let li = $('<li/>', { class: 'nav-item'}).appendTo(this.toc);
```
piotrekpasciak: Are these 3 `let` mutated anywhere? I think we could use `const` instead.

nemaniarjun: Will fix!

nemaniarjun: I kept them as `let` as I thought `const` wont work inside a `.each()` block, but it does. 

Thanks!

app/javascript/stimulus_js/controllers/challenge_overview_controller.js
nemaniarjun: aaah typos :/ nice catch!

Issue #960, opened 2019-06-01
Receive error when user has no posts


seanfcarroll:
When I click on "challenges" or "bio" in my profile, I see the "sean has not provided any information yet. " message".

When I click on "posts" I see an error page. (notice also the double underline).

<img width="472" alt="Screenshot 2019-06-01 at 16 43 28" src="https://user-images.githubusercontent.com/11340230/58749967-9749e100-848c-11e9-8224-f5c6c3bef14d.png">

<img width="813" alt="Screenshot 2019-06-01 at 16 43 48" src="https://user-images.githubusercontent.com/11340230/58749975-9ca72b80-848c-11e9-95fe-fb6acb0b4212.png">

nemaniarjun:
Okay, so this is supposed to redirect to our discourse's user activity page. 
The error is due to differing ids on rails and discourse.

The underline is a small issue, we are anyway redirecting the user away to discourse.

Issue #832, opened 2019-02-17
Fetch details of user from discourse for participant page.


nemaniarjun:

nemaniarjun:
This was made to add the posts made by a user on discourse, to the profile on AIcrowd.
@spMohanty do we need this?

Ref: https://aicrowd-design.netlify.com/template-user-posts

spMohanty:
Yes, we still do need this. Lets see if theres an easy way to do this using the Discourse  APIs.

nemaniarjun:
Cant find any way to do this on SO and the docs.

spMohanty:
@nemaniarjun : Maybe https://github.com/discourse/discourse_api ?

nemaniarjun:
# Get Comments
https://discourse.aicrowd.com/user_actions.json?username=arjun_nemani&filter=5
# Get Posts and Comments
https://discourse.aicrowd.com/user_actions.json?username=arjun_nemani&filter=4,5

nemaniarjun:
Do we show only the posts? or Both? @jsnrynlds Can you check the design as per the jsons above?

jsnrynlds:
Good question @nemaniarjun. I think we can show only the comments/posts but with the topic title above.

Will this be styled as per [template designs](https://aicrowd-design.netlify.com/template-user-posts) or using Discourse?

nemaniarjun:
We would need to come up with the design.

PR #1193, opened 2020-03-03
Display Discourse posts on Posts tab

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/474972252

### DIFF
--- a/app/services/discourse/update_category_service.rb
+++ b/app/services/discourse/update_category_service.rb
@@ -8,7 +8,7 @@ module Discourse
     def call
       retry_count ||= 0
 
-      return failure('DiscourseApi client couldn\'t be properly initialized.') if client.nil?
+      return failure('Discourse API client couldn\'t be properly initialized.') if client.nil?
       return failure('Discourse category doesn\'t exist in our database.') if challenge.discourse_category_id.blank?
 
       update_category_request(ensure_uniqueness: retry_count.positive?)
@@ -37,9 +37,8 @@ module Discourse
       # - category name has to be unique
       # - slug has to be unique
       client.put(
-        "/categories/#{challenge.discourse_category_id}", {
+        "/categories/#{challenge.discourse_category_id}.json", {
           name:       truncated_string(challenge.challenge, ensure_uniqueness),
-          slug:       truncated_string(challenge.slug, ensure_uniqueness),
           color:      ::Discourse::BaseService::CATEGORY_DEFAULT_COLOR,
           text_color: ::Discourse::BaseService::CATEGORY_DEFAULT_TEXT_COLOR
         }
--- a/app/views/participants/_discourse_posts_list.html.erb
+++ b/app/views/participants/_discourse_posts_list.html.erb
@@ -1,0 +1,53 @@
+<section class="section-p-md">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-12">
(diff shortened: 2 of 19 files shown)


nemaniarjun:
Closes https://github.com/AIcrowd/AIcrowd/issues/832

nemaniarjun:
Should fix https://github.com/AIcrowd/AIcrowd/issues/960

app/services/discourse/update_category_service.rb
```diff
@@ -37,7 +37,7 @@ def update_category_request(ensure_uniqueness: false)
       # - category name has to be unique
       # - slug has to be unique
       client.put(
-        "/categories/#{challenge.discourse_category_id}", {
+        "/categories/#{challenge.discourse_category_id}.json", {
           name:       truncated_string(challenge.challenge, ensure_uniqueness),
```
nemaniarjun: Do we need to still truncate the string? Does discourse automatically do it if it exceeds the limits? Should we set the slug ourselves? or just set the name and fetch the slug that discourse assigns ? That way we don't have to check for uniqueness and discourse will be free to manage the slugs with more freedom.

app/views/participants/_discourse_posts_list.html.erb
```diff
@@ -0,0 +1,53 @@
...
+                  <!-- topic – comment -->
+                  <div class="topic-comment">
+                    <%= link_to participant_path(participant) do %>
+                      <%= image_tag participant.image_url, class: 'avatar' %>
+                    <% end %>
+                    <div class="discussion-topic-body">
+                      <h3><%= link_to discourse_post['topic_title'], discourse_topic_url(discourse_post['topic_slug']), target: '_blank' %></h3>
+                      <span class="discussion-details"><%= discourse_post_datetime(discourse_post['created_at']) %></span>
+
+                      <p><%= sanitize_html(discourse_post['cooked']) %></p>
```
nemaniarjun: :100: Finally we are using discourse properly!!!! :tada:

spec/vcr/discourse_api/data_explorer_queries/user_posts/success.yml
```diff
@@ -0,0 +1,101 @@
+---
```
nemaniarjun: VCR is great!!!!

app/services/discourse/update_category_service.rb
piotrekpasciak: Yes, we still need to truncate the string. Discourse has multiple validations that are listed in comments before this request. If we don't send data that complies with Discourse validations, request will fail.

piotrekpasciak: We can't allow users to choose the slugs as it may fail validation. Discourse doesn't return slug. proposition - it only returns list of errors describing what is wrong with slug.

nemaniarjun: According to the developer documentation for this https://docs.discourse.org/#tag/Categories/paths/~1categories.json/get
We only need to send name, color and text-color. I still think we should not be setting the slug and let discourse do it for us if it can, can you check by making a request manually?

piotrekpasciak: Okey, I'll doublecheck.

piotrekpasciak: Discourse API seems to handle slug uniqueness. I've updated the PR.

Issue #19, opened 2018-12-04
Define challenge categories/tags


spMohanty:
_From @spMohanty on April 26, 2018 23:41_

We could mark every challenge with categories : Educational Challenge (simple challenges on popular datasets like MNIST, etc), Academic Challenge (NIPS Challenges, etc), Challenges with Prizes, Challenges with Academic papers, Challenges with Travel Grants, etc. 

What I imagine are small icons next to challenge names (on the home page, and also on the challenge page), which gives a broad view of what type of challenges are out there.
Could help participants to prioritize their efforts depending on what they are after.

Needs some thought and design by @jsnrynlds too.

_Copied from original issue: crowdAI/crowdai#726_

PR #1195, opened 2020-03-04
Category tags for challenge

alkeshrails:
Category/Tags for the challenges
link : https://aicrowd.monday.com/boards/363093650/pulses/465361490

### DIFF
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -3,6 +3,23 @@ $(document).on('click', '.challenge-invite-participant', function(){
   $('.challenge-invite-participant').attr("disabled", true);
 });
 
+$(document).on('click', '.category_opt', function(){
+  var new_value = $('#challenge_selected_category').val().replace($(this).val(), '');
+  $('#challenge_selected_category').val(new_value);
+  $(this).removeClass('badge badge-success category_opt');
+  $('#challenge_category').append($(this));
+});
+
+$(document).on('change', '#challenge_category', function(){
+  var new_value = $('#challenge_selected_category').val() + ' ' + $(this).val();
+  $('#challenge_selected_category').val(new_value);
+  $('.show_category').append($("option:selected", this));
+  $(".show_category option").each(function(){
+    var text = $(this).val();
+    $(this).addClass('badge badge-success category_opt').attr('id', text);
+  });
+});
+
 function resetChallengesFormClientValidations() {
   // We need to wait till fields show up in browser
   setTimeout(function() { $('#challenges-form').enableClientSideValidations(); }, 1000);
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -3,21 +3,8 @@ $(document).on('click', '.challenge-invite-participant', function(){
   $('.challenge-invite-participant').attr("disabled", true);
 });
 
-$(document).on('click', '.category_opt', function(){
-  var new_value = $('#challenge_selected_category').val().replace($(this).val(), '');
-  $('#challenge_selected_category').val(new_value);
-  $(this).removeClass('badge badge-success category_opt');
-  $('#challenge_category').append($(this));
-});
-
-$(document).on('change', '#challenge_category', function(){
-  var new_value = $('#challenge_selected_category').val() + ' ' + $(this).val();
-  $('#challenge_selected_category').val(new_value);
-  $('.show_category').append($("option:selected", this));
-  $(".show_category option").each(function(){
-    var text = $(this).val();
-    $(this).addClass('badge badge-success category_opt').attr('id', text);
-  });
+$(document).ready(function() {
+  $('.category_select2').select2();
 });
 
 function resetChallengesFormClientValidations() {
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -188,7 +188,7 @@ class ChallengesController < ApplicationController
 
   def set_category
     category = @challenge&.categories.present? ? Category.where.not(id: @challenge.categories.ids) : Category.all
-    category.map{|c| [c.name, c.id]}
+    @categories = category.map{|c| [c.name, c.id]}
   end
 
   def challenge_params
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -9,6 +9,7 @@ class ChallengesController < ApplicationController
   before_action :set_organizer, only: [:edit, :update]
   before_action :set_organizers_for_select, only: [:new, :create]
   before_action :set_challenge_rounds, only: [:edit, :update]
+  before_action :set_category, only: [:new, :create , :edit]
 
   respond_to :html, :js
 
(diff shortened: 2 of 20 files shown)


app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -3,6 +3,23 @@ $(document).on('click', '.challenge-invite-participant', function(){
   $('.challenge-invite-participant').attr("disabled", true);
 });
 
+$(document).on('click', '.category_opt', function(){
```
piotrekpasciak: On which pages do you want to run these 2 events? Do you think we could put them inside Paloma controller action, so this code wouldn't be run everywhere?

app/controllers/challenges_controller.rb
```diff
@@ -174,6 +178,18 @@ def create_invitations
     end
   end
 
+  def create_challenge_category
+    @challenge.category_challenges.destroy_all if @challenge.category_challenges.present?
```
piotrekpasciak: This line doesn't make mutch sense. You shouldn't need to delete all categories from challenge before creating new.

db/migrate/20200302110003_create_categories.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCategories < ActiveRecord::Migration[5.2]
+  def change
+    create_table :categories do |t|
+      t.string :name
```
piotrekpasciak: It would be good idea to add `null: fase` constraint here and unique index. Add related validations to model as well.

db/migrate/20200302110125_create_category_challenges.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCategoryChallenges < ActiveRecord::Migration[5.2]
+  def change
+    create_table :category_challenges do |t|
+      t.references :category
```
piotrekpasciak: It would be good idea to add `null: false` constraints here for both challenge and category. Add corresponding validations to model as well.

spec/factories/category_challenges.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :category_challenge do
+    category_id { "" }
```
piotrekpasciak: `association(:category)` and  `association(:challenge)` would make more sense here.

spec/models/category_challenge_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe CategoryChallenge, type: :model do
```
piotrekpasciak: Remove empty spec files from repository.

app/controllers/challenges_controller.rb
```diff
@@ -174,6 +178,18 @@ def create_invitations
...
+    params[:challenge][:selected_category].split(' ').each do |category|
+      @challenge.category_challenges.create(category_id: category)
+    end
+  end
+
+  def set_category
+    category = @challenge&.categories.present? ? Category.where.not(id: @challenge.categories.ids) : Category.all
+    @categories ||= category.collect{|c| [c.name, c.id]}
```
piotrekpasciak: `collect` is deprecated - you should use map instead:
https://github.com/rubocop-hq/ruby-style-guide#map-find-select-reduce-size

app/controllers/challenges_controller.rb
```diff
@@ -174,6 +178,18 @@ def create_invitations
     end
   end
 
+  def create_challenge_category
+    @challenge.category_challenges.destroy_all if @challenge.category_challenges.present?
+    params[:challenge][:selected_category].split(' ').each do |category|
```
piotrekpasciak: Why did you choose this approach to create  categories? It's impossible to use nested attributes?
https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html

app/controllers/challenges_controller.rb
piotrekpasciak: Memoization doesn't make sense here are this method is called only once per request.

app/assets/javascripts/controllers/challenges_controller.js
alkeshrails: I want to run this on page https://github.com/AIcrowd/AIcrowd/blob/4fd06a51ebaad2b8f5b56052f98a0ae98ced950e/app/views/challenges/tabs/_details_form.html.erb
So I added on challenge_controller js instead of on partial page 'details_form' partial page.

app/controllers/challenges_controller.rb
alkeshrails: Actually If I remove a category to edit page in that case, It remove all categories and create with existing category_ids. We don't have ajax for remove category.Because category add and remove is only done when we click update button for challenge.

db/migrate/20200302110003_create_categories.rb
alkeshrails: Sure, will do.

db/migrate/20200302110125_create_category_challenges.rb
alkeshrails: thanks

spec/factories/category_challenges.rb
alkeshrails: Thanks.

spec/models/category_challenge_spec.rb
alkeshrails: Sure.

app/controllers/challenges_controller.rb
alkeshrails: Sure!

app/controllers/challenges_controller.rb
alkeshrails: If I use nested attribute, than I need to add `add category` that will show multiple select box of categories on UI, it would be ugly. So I use this approach. User can select multiple categories to a select box and record create.

app/controllers/challenges_controller.rb
alkeshrails: Thanks

app/assets/javascripts/controllers/challenges_controller.js
piotrekpasciak: Okey, let's keep it as it is, as this partial is rendered by both new and edit page.

app/controllers/challenges_controller.rb
piotrekpasciak: I'll take a look on it tomorrow and I let you know if it can be done with nested attributes without to much effort.

nemaniarjun:
Closes https://github.com/AIcrowd/AIcrowd/issues/19

db/migrate/20200302110003_create_categories.rb
alkeshrails: Done

db/migrate/20200302110125_create_category_challenges.rb
alkeshrails: Done

spec/factories/category_challenges.rb
alkeshrails: Done

app/controllers/challenges_controller.rb
alkeshrails: Done

alkeshrails: done

app/models/category_challenge.rb
```diff
@@ -0,0 +1,6 @@
+class CategoryChallenge < ApplicationRecord
+  belongs_to :category
+  belongs_to :challenge
+  validates :challenge_id, uniqueness: { scope: :category_id }
+  validates :challenge_id, :category_id, presence: true
```
piotrekpasciak: This line should be removed, `belongs_to` has provides presence validation for association since Rails 5.0.

PR #1198, opened 2020-03-04
 Challenge Insights: Add charts for insights

nemaniarjun:
https://aicrowd.monday.com/boards/363093650/pulses/476823993

Rake Task: `rake ahoy_visits:update_location `


app/controllers/insights_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+                    # sort_order == "ascending" or "not_used"
+                    get_running_agg('max', grouped_collection.minimum(:score), precision)
+                  end
+
+    render json: return_hash
+  end
+
+  def challenge_participants_country
+    country_count = Hash.new(0)
+    participants  = @challenge.challenge_participants.map(&:participant).select(&:present?)
```
piotrekpasciak: It would be better to add `has_many :participants through: challenge_participants` relation and change this line to something like `@challenge.participants`.

app/controllers/insights_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    render json: country_count.except(nil)
+  end
+
+  private
+
+  def get_running_agg(fn, grouped_collection, precision)
+    ret       = {}
+    current   = 0
+    current   = 100000000000000000000 if fn == 'min'
```
piotrekpasciak: ` Float::INFINITY` might be a better choice then this number.

spec/features/insights_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+describe InsightsController, feature: true do
```
piotrekpasciak: 👍

app/controllers/insights_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    render json: return_hash
+  end
+
+  def challenge_participants_country
+    country_count = Hash.new(0)
+    participants  = @challenge.challenge_participants.map(&:participant).select(&:present?)
+    participants.each do |participant|
+      country                 = ISO3166::Country[participant.country_cd]&.name
+      country               ||= participant.visits.where.not(country: nil).first&.country
```
skbly7: `.last` might make more sense here, so we update the graphs in case users moved to newer countries automatically. 😄

app/views/insights/_charts/_challenge_participants_country.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="chart-wrapper pb-5" data-controller="chart" data-chart-index="3" data-chart-chart-type="GeoChart">
+  <header class="section-header-decorative">
+    <div>
+      <h4>Geo Chart</h4>
```
skbly7: Add this CSS:
```
.chart-wrapper h4 {
    text-transform: none;
}
```

PR #1201, opened 2020-03-05
Add Discourse::ApiClient class and remove discourse_api gem

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/482076517
https://aicrowd.monday.com/boards/363093650/pulses/480528360


spec/rails_helper.rb
```diff
@@ -9,6 +9,9 @@
 require 'support/config/warden'
 require 'support/config/vcr'
 
+# Load Error classes from Discourse::ApiClient
+require "#{Rails.root}/app/services/discourse/api_client.rb"
```
skbly7: (Minor) Is this the correct way to add or is there any better way to do so?

piotrekpasciak: Standard Ruby on Rails approach is to add file_name corresponding to defined class (file will be loaded to memory whenever class has been referenced), but adding a separate file for each error class is overkill. Unfortunetly I didn't find a better approach.

PR #1208, opened 2020-03-06
Fix category badge UI on show and index page

alkeshrails:


app/views/shared/challenges/_listing.html.erb
```diff
@@ -33,7 +33,9 @@
         </div>
 
         <!-- prizes breakdown -->
-        <%= render partial: 'shared/challenges/prizes_breakdown', locals: { challenge: challenge } %>
+        <div class="prizes-breakdown">
```
skbly7: This <div> would be good to have it inside the partial itself, but just outside if/else condition? So code in listing.html.erb is compact?

alkeshrails: Okay, doing!

PR #1211, opened 2020-03-06
Hotfix/Best Score vs Time 

nemaniarjun:


lib/tasks/ahoy_visits.rake
```diff
@@ -1,8 +1,25 @@
...
+      next if visit.country.present?
+
+      puts "Processing #{visit.id}"
       Ahoy::GeocodeJob.perform_now(visit)
+      puts "Processed #{visit.id}"
+    end
+  end
+
+  desc "Detect location for participants"
+  task update_participant_location: :environment do
```
skbly7: How can we do this more new signups? 

Can we create this as function in Participants model instead:

```
def deletect_country_if_not_present
      return if participant.country_cd.present?

      detected_country = participant.visits.where.not(country: nil)&.last&.country
      next unless detected_country.present?

      country_code = ISO3166::Country.all_names_with_codes.to_h[detected_country]
      participant.update!(country_cd: country_code)
```

And call that function in this rake task? And call this function on email confirmation?

Issue #1215, opened 2020-03-09
"View Submissions" button on leaderboard is not consistent


spMohanty:
Some submissions on the leaderboard do not have a "View Submissions" button.

<img width="1292" alt="Screenshot 2020-03-09 at 10 29 21" src="https://user-images.githubusercontent.com/1581312/76199916-00ffd400-61f1-11ea-9b87-bc82de7f0a32.png">

piotrekpasciak:
@spMohanty @shivam This is already deployed to production

PR #1217, opened 2020-03-09
Fix View button in leaderboards listing

piotrekpasciak:
Github issue:
https://github.com/AIcrowd/AIcrowd/issues/1215

### DIFF
--- a/app/views/leaderboards/_leaderboards.html.erb
+++ b/app/views/leaderboards/_leaderboards.html.erb
@@ -1,8 +1,7 @@
 <% leaderboards.each do |leaderboard| %>
   <%
-    participants            = leaderboard_participants(leaderboard)
-    challenge               = leaderboard.challenge
-    leaderboard_participant = leaderboard.participant
+    participants = leaderboard_participants(leaderboard)
+    challenge    = leaderboard.challenge
   %>
 
   <tr>
@@ -53,10 +52,10 @@
 
     <td class="text-right"><%= local_time(leaderboard.created_at, "%a, %e %b %Y %H:%M") %></td>
     <td class="text-right">
-      <% if leaderboard_participant.present? %>
-        <%= link_to challenge_submission_path(challenge, leaderboard.submission_id), class: 'btn btn-secondary btn-sm' do %>
-          View
-        <% end %>
+      <% if leaderboard.submitter_type != 'OldParticipant' && leaderboard.submitter.present? && leaderboard.submission.present? %>
+        <%= link_to 'View', challenge_submission_path(challenge, leaderboard.submission_id), class: 'btn btn-secondary btn-sm' %>
+      <% else %>
+        <%= link_to 'View', '#', class: 'btn btn-secondary btn-sm disabled disabled--with-pointer-events', title: 'Submitter or Submission was deleted' %>
       <% end %>
     </td>
   </tr>
--- a/app/assets/stylesheets/application-redesign.css.scss
+++ b/app/assets/stylesheets/application-redesign.css.scss
@@ -2733,6 +2733,11 @@ fieldset:disabled a.btn {
   pointer-events: none;
 }
 
+a.btn.disabled--with-pointer-events,
+fieldset:disabled a.btn {
+  pointer-events: initial;
+}
+
 .btn-primary {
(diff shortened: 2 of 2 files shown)


app/views/leaderboards/_leaderboards.html.erb
```diff
@@ -53,10 +52,11 @@
...
     <td class="text-right">
-      <% if leaderboard_participant.present? %>
-        <%= link_to challenge_submission_path(challenge, leaderboard.submission_id), class: 'btn btn-secondary btn-sm' do %>
-          View
-        <% end %>
+      <% if leaderboard.submitter.present? && leaderboard.submission.present? %>
+        <%= link_to 'View', challenge_submission_path(challenge, leaderboard.submission_id), class: 'btn btn-secondary btn-sm' %>
+      <% else %>
+        <!-- Submitter or Submission was deleted -->
+        <%= link_to 'View', '#', class: 'btn btn-secondary btn-sm disabled', disabled: true %>
```
nemaniarjun: Please check the [team helpers](https://github.com/AIcrowd/AIcrowd/blob/737f604ac807b86476e9d62f1703114303b2eae8/app/helpers/team_helper.rb) on how we can set tooltip text on the disabled button. 
AFAI remember you just need to set html title attrib

alkeshrails: Agree with @nemaniarjun , we should show title on disabled field.

piotrekpasciak: @nemaniarjun @alkeshrails I've added this comment as HTML 5 title tooltip.

app/views/leaderboards/_leaderboards.html.erb
```diff
@@ -53,10 +52,10 @@
 
     <td class="text-right"><%= local_time(leaderboard.created_at, "%a, %e %b %Y %H:%M") %></td>
     <td class="text-right">
-      <% if leaderboard_participant.present? %>
-        <%= link_to challenge_submission_path(challenge, leaderboard.submission_id), class: 'btn btn-secondary btn-sm' do %>
-          View
-        <% end %>
+      <% if leaderboard.submitter.present? && leaderboard.submission.present? %>
```
skbly7: This will break for UnknownParticipant leaderboards afaik. @nemaniarjun?

skbly7: Tested for UnknownParticipant and this thing will break:

image


```

      <% if participants.empty? %>	      <% if participants.empty? %>
        <%= image_tag default_image_url, class: 'avatar' %>	        <%= image_tag default_image_url, class: 'avatar' %>
        <span class="text-link">Unknown User</span>	        <span class="text-link">Unknown User</span>
      <% elsif participants.length == 1 %>	      <% elsif participants.length == 1 %>
        <%= link_to participant_path(participants[0]), title: participants[0].name, data: {toggle: :tooltip} do %>	        <%= link_to participant_path(participants[0]), title: participants[0].name, data: {toggle: :tooltip} do %>
          <%= image_tag participants[0].image_url, class: 'avatar' %>	          <%= image_tag participants[0].image_url, class: 'avatar' %>
          <span class="text-link"><%= participants[0].name %></span>	          <span class="text-link"><%= participants[0].name %></span>
        <% end %>	        <% end %>
      <% else %>	      <% else %>
```

^ Check line number 14 of this file, might make sense to do like that.

skbly7: Still will have the same error as described above, the if condition thing. Please go again from traceback screenshot above.

piotrekpasciak: @skbly7 Is there a URL address where I can replicate that error? I'm not sure how to replicate.

piotrekpasciak: I think the bug is a result of that some records on production have `submitter_type` of value: OldParticipant and we don't even have class like that.

piotrekpasciak: @skbly7 I've fixed issue described above, by additional if condition:
`leaderboard.submitter_type != 'OldParticipant'`

Fixing this issue properly would require major refactoring, thus I've decided to use fix above.

PR #1221, opened 2020-03-10
Fix nil exception on @current_round in leaderboards#index

piotrekpasciak:
Rollbar: 
https://rollbar.com/aicrowd-production/aicrowd-production/items/276/

### DIFF
--- a/app/views/leaderboards/index.html.erb
+++ b/app/views/leaderboards/index.html.erb
@@ -44,8 +44,8 @@
                 <% if @challenge.media_on_leaderboard %>
                   <th scope="col">Media</th>
                 <% end %>
-                <th scope="col"><%= @current_round.get_score_title %></th>
-                <% if @current_round.get_score_secondary_title.present? %>
+                <th scope="col"><%= @current_round&.get_score_title %></th>
+                <% if @current_round&.get_score_secondary_title.present? %>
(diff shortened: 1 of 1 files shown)


app/views/leaderboards/index.html.erb
```diff
@@ -44,8 +44,8 @@
                 <% if @challenge.media_on_leaderboard %>
                   <th scope="col">Media</th>
                 <% end %>
-                <th scope="col"><%= @current_round.get_score_title %></th>
-                <% if @current_round.get_score_secondary_title.present? %>
+                <th scope="col"><%= @current_round&.get_score_title %></th>
```
nemaniarjun: I am a bit concered about this function. I remember creating it but also thinking this is a weird place to be doing things like this. @skbly7 lets have a proper selection on the challenge edit page with "number of scores to use" could be anything from 1 score upto to 7 (current limit) and we can refactor the hacky code I wrote for multi-score support

skbly7: Yup, scoring thing and leaderboard in general needs to be improved. And this is something which should be present when we improve it.

I remember we had 2-3 calls discussion for just the leaderboard in the past, do you have doc of it? If ye, please share it in rails channel on Slack.

Issue #1220, opened 2020-03-09
Add a hashtag to category rendering ?


spMohanty:
Should we add a hash to the tags, so that they are more apparent as tags/categories ?

For example : "#reinforcement-learning" instead of "reinforcement-learning", we can probably do this only during the rendering. But from the UX point of view, its much more apparent then that this is a tag.

Thoughts @skbly7  ?

skbly7:
Sure, I think that makes sense. We were considering the feature in-middle of "Category" / "Tags" causing not to go ahead with hashtag thing earlier.

But based on developments over the last few days, I feel more comfortable treating them as tags and have UX modifications accordingly. I will include it in category design improvement discussion.

piotrekpasciak:
@skbly7 @spMohanty 
This looks to be already deployed to production.

PR #1222, opened 2020-03-10
Feature/filter for the challenges

alkeshrails:
Filters for the challenges (running, ended, category)
link : https://aicrowd.monday.com/boards/363093650/pulses/465361431


app/views/challenges/index.js.erb
```diff
@@ -1,5 +1,5 @@
 $(".load-more").remove();
-$("#challenges-div").append("<%= j render partial: 'shared/challenges/listing', locals: { challenges: @challenges } %>");
+$("#challenges-div").html("<%= j render partial: 'shared/challenges/listing', locals: { challenges: @challenges } %>");
```
nemaniarjun: I am concerned about this line. Does "load more" work after this change?  
It will just replace the challenges instead of loading more.

app/views/landing_page/index.js.erb
```diff
@@ -0,0 +1 @@
+$(".challenge-list").html("<%= j render partial: 'shared/challenges/listing', locals: { challenges: @challenges } %>");
```
nemaniarjun: Do we need this like this? Can we put this in landing_page view? I have trouble understanding when/how is this js triggered?

app/views/shared/challenges/_filter.html.erb
```diff
@@ -0,0 +1,43 @@
+<%= form_tag path_of_index_page, method: :get, remote: true, class: 'filter-dropdown-form' do %>
```
nemaniarjun: I am guessing this form-submit is triggering the landing_page index.js.erb.

app/controllers/concerns/challenge_filter.rb
```diff
@@ -0,0 +1,34 @@
+module ChallengeFilter
+  extend ActiveSupport::Concern
```
nemaniarjun: I have read case study against the use of `concers`, maybe we should make this into a service object instead? Or are service objects for external APIs only ? cc @piotrekpasciak @skbly7

app/controllers/concerns/challenge_filter.rb
```diff
@@ -0,0 +1,34 @@
...
+  def set_filters
+    @categories = Category.all
+    @status     = Challenge.statuses.keys
+    @prize_hash = { prize_cash:     'Cash prizes',
+                    prize_travel:   'Travel grants',
+                    prize_academic: 'Academic papers',
+                    prize_misc:     'Misc prizes' }
+  end
+
+  def filter_challenge(challenges = nil)
```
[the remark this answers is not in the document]
nemaniarjun: This part of the code is very complex we should try to simplify it a bit if we can.

app/controllers/concerns/challenge_filter.rb
```diff
@@ -0,0 +1,34 @@
...
+    challenges = category? ? Challenge.joins(:categories).where('categories.id IN (?)', params[:category]['category_ids']) : all_list(challenges)
+    challenges = challenges.where(status_cd: params[:status]) if params[:status].present?
+    if params[:prize].present?
+      challenges = challenges.where.not(prize_cash: nil) if params[:prize]['prize_type'].include?('prize_cash')
+      challenges = challenges.where.not(prize_travel: nil) if params[:prize]['prize_type'].include?('prize_travel')
+      challenges = challenges.where.not(prize_academic: nil) if params[:prize]['prize_type'].include?('prize_academic')
+      challenges = challenges.where.not(prize_misc: nil) if params[:prize]['prize_type'].include?('prize_misc')
```
nemaniarjun: how about something like
```python
for each prize_type in params[:prize]: 
   next if prize_type not in @prize_hash.keys
   challenge.where.not(prize_type: nil)
```

app/controllers/concerns/challenge_filter.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+    challenges
+  end
+
+  private
+
+  def category?
+    params[:category] && params[:category]['category_ids'].present?
```
nemaniarjun: use the safe navigation operator and [dig](https://sts10.github.io/2016/01/07/using-ruby-2-dot-3-s-dig-method-to-navigate-the-facebook-insights-object.html) for cleaner code! 
something like: 
`params.dig(:category, 'category_ids').present? `

app/controllers/landing_page_controller.rb
```diff
@@ -1,8 +1,14 @@
 class LandingPageController < ApplicationController
+  include ChallengeFilter
```
nemaniarjun: Can we keep this in challenge index controller only using the form & js call. 
Idea: Landing Page view has the a form with url set to challenges#index or some other load_challenges endpoint which returns filtered challenges.
The same endpoint could be used directly from the challenges index page with some changes for load-more to work properly.

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -5,6 +5,24 @@ $(document).on('click', '.challenge-invite-participant', function(){
...
+    $('#challenge-statuses').addClass('filter-active');
+  });
+  $('input[name="category[category_ids][]"]:checkbox').change(function(){
+    $('#challenge-categories').removeClass('filter-active');
+    if ($('input[name="category[category_ids][]"').is(':checked'))
+    {
+      $('#challenge-categories').addClass('filter-active');
+    }
+  });
+  $('input[name="prize[prize_type][]"]:checkbox').change(function(){
```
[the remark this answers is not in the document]
nemaniarjun: this is really identical we should be able to roll it into one loop

app/controllers/concerns/challenge_filter.rb
alkeshrails: Service is used for external API and here we use this in 2 controller so I make concern.

app/controllers/concerns/challenge_filter.rb
alkeshrails: All conditions is independent that's why its looking complex.

app/controllers/landing_page_controller.rb
alkeshrails: I think for this we need to discuss with team. Because currently we have 2 view for challenges. One is Landing page index its show Challenges except draft status. And on challenge index it show all.

app/views/challenges/index.js.erb
alkeshrails: Actually If I append search result on index page It will show with other not searched challenges. So user will be confuse here. So I used html.
And yes, .load-more is removing.

app/views/landing_page/index.js.erb
alkeshrails: when user apply filter on Landing index page. so Instead of reload page I make JS call and show filtered result in .challenge-list

app/views/shared/challenges/_filter.html.erb
alkeshrails: This form is triggering from landing#index and challenge#index.

app/controllers/concerns/challenge_filter.rb
piotrekpasciak: ServiceObject is a general-purpose pattern that can be used to implement any kind of business logic - it's Ruby implemenatation of classical Command pattern. There is a bad vibe about Concerns in Ruby on Rails community, thus from my point of view it's not recommended approach.

I would agree with @nemaniarjun that it would be more beneficial to implement `filter_challenge` as ServiceObject class with test coverage. You can call this class something like `Challenges::FilterService`.

I would put `set_filters` method in both controllers. It's small duplication, but shouldn't be a big deal.

app/controllers/concerns/challenge_filter.rb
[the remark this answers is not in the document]
piotrekpasciak: `https://github.com/AIcrowd/AIcrowd/pull/1222/files#diff-ffc3bf214aaabc96b5563912960ae637R15`` - this could be moved to private method.  Not sure if we can do mutch more about this method/

app/views/challenges/index.js.erb
piotrekpasciak: I agree @nemaniarjun we need to test if filtering and load more functionality works well together.

app/views/shared/challenges/_filter.html.erb
```diff
@@ -0,0 +1,43 @@
...
+      <% @prize_hash.each do |p_key, p_value| %>
+        <div class="form-check">
+          <%= check_box_tag 'prize[prize_type][]', p_key, false, {class: 'form-check-input'} %>
+          <%= label_tag "prize[prize_type][][#{p_key}]", p_value, class: 'form-check-label' %>
+        </div>
+      <% end %>
+      <hr>
+    </div>
+  </div>
+  <%= submit_tag 'Apply', class: 'btn btn-secondary btn-sm' %>
```
skbly7: It doesn't look good.

Let's put it as <a> only instead of big button.

Current UI

image

Suggested for now:

image

```
<a style="color: #F0524D;margin-left: 1rem;">
    <span title="Apply filter">Apply</span>
</a>
```

app/views/shared/challenges/_filter.html.erb
```diff
@@ -0,0 +1,43 @@
...
+          </label>
+        </div>
+      <% end %>
+    </div>
+  </div>
+  <div class="dropdown">
+    <div class="filter dropdown-toggle" id="challenge-prizes" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      Prizes
+    </div>
+    <div class='dropdown-menu form-dropdown'>
```
skbly7: Prizes thing has a weird horizontal line/seperator coming, I guess it isn't removed properly.

image

app/views/landing_page/index.js.erb
skbly7: The landing page don't need filters. Please remove it.

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -5,6 +5,24 @@ $(document).on('click', '.challenge-invite-participant', function(){
```
skbly7: We discussed that the links need to be sharable after the filters are applied and so on.

Currently, the page URL doesn't change when filter are active v/s not and so on.

Example:
https://aicrowd-feature-filter-eilxalm.herokuapp.com/challenges

On filter applied
https://aicrowd-feature-filter-eilxalm.herokuapp.com/challenges?status=completed&category%5Bcategory_ids%5D%5B%5D=1

^ side note, I will prefer in case we can show category name instead of ID in the URL, they will be more appealing when shared later in future, plus hashtag aspect.

app/services/challenge_filter_service.rb
```diff
@@ -0,0 +1,27 @@
+class ChallengeFilterService
```
piotrekpasciak: You should put this class under `Challenges` namespace, we need to group ServiceObject classes, otherwise, it will be difficult to manage them in future, when we'll have a lot of them.

app/services/challenge_filter_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(params, challenges)
+    @params = params
+    @challenges = challenges
+  end
+
+  def filter_challenge
```
piotrekpasciak: ServiceObject pattern should implement only constructor and single public method called `call`. Please rename that method to `call`.

app/services/challenge_filter_service.rb
piotrekpasciak: Please write a unit test coverage for this class. Write at least one spec that covers positive path of this class.

app/controllers/challenges_controller.rb
```diff
@@ -176,6 +178,19 @@ def set_challenge_rounds
...
+    @categories = Category.all
+    @status     = challenge_status
+    @prize_hash = { prize_cash:     'Cash prizes',
+                    prize_travel:   'Travel grants',
+                    prize_academic: 'Academic papers',
+                    prize_misc:     'Misc prizes' }
+  end
+
+  def challenge_status
+    params[:controller].eql?("landing_page") ? Challenge.statuses.keys - ['draft'] : Challenge.statuses.keys
```
piotrekpasciak: Why do you use `eql?` method here? You don't need to compare here both value and class as you use it against 2 strings. Please use `==` operator.

skbly7:
Looks good now. Changes requested:

1. Based on URL params those values can be pre-selected. Example: status=running in param can reflect back in filters.
2. Separate PR which can be taken now: Making your category feature now integrated with filters. i.e. making those rendering as clickable (+ prefix hashtag #1220).

spec/services/challenges/filter_service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:category_challenge) { create(:category_challenge, challenge: challenge1, category: category) }
+    let(:params) { create(:category_challenge, challenge: challenge1, category: category) }
+
+    context 'when categories name, status and prize cash are provided' do
+      it 'return challenges as per given fields' do
+        params = ActionController::Parameters.new(category: { category_names: [category.name] }, status: challenge1.status_cd, prize: { prize_type: challenge1.prize_cash })
+        challenge1
+        challenge2
+        category_challenge
+        subject = described_class.new(params, Challenge.all)
```
piotrekpasciak: `subject` declaration should be on top of the class e.g:
https://github.com/AIcrowd/AIcrowd/blob/master/spec/services/challenges/import_service_spec.rb#L4

spec/services/challenges/filter_service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:challenge1) { create(:challenge) }
+    let(:challenge2) { create(:challenge) }
+    let(:category) { create(:category) }
+    let(:category_challenge) { create(:category_challenge, challenge: challenge1, category: category) }
+    let(:params) { create(:category_challenge, challenge: challenge1, category: category) }
+
+    context 'when categories name, status and prize cash are provided' do
+      it 'return challenges as per given fields' do
+        params = ActionController::Parameters.new(category: { category_names: [category.name] }, status: challenge1.status_cd, prize: { prize_type: challenge1.prize_cash })
+        challenge1
```
piotrekpasciak: Use `let!` instead of https://github.com/AIcrowd/AIcrowd/pull/1222/files#diff-b02fb7e87a895466830d90b95aba947aR15-R17

`let!` will ensure that data is created. before the test.

spec/services/challenges/filter_service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  describe '#call' do
+    let(:challenge1) { create(:challenge) }
+    let(:challenge2) { create(:challenge) }
+    let(:category) { create(:category) }
+    let(:category_challenge) { create(:category_challenge, challenge: challenge1, category: category) }
+    let(:params) { create(:category_challenge, challenge: challenge1, category: category) }
+
+    context 'when categories name, status and prize cash are provided' do
+      it 'return challenges as per given fields' do
+        params = ActionController::Parameters.new(category: { category_names: [category.name] }, status: challenge1.status_cd, prize: { prize_type: challenge1.prize_cash })
```
piotrekpasciak: You declare `params` two times in this test. Once by `let` and once. by local variable. Using let is preferred approach.

PR #1223, opened 2020-03-10
Display Discourse posts on challenge Discussion tab

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/476851676


config/routes.rb
```diff
@@ -158,7 +158,9 @@
     end
   end
 
-
+  resources :challenges, only: [], module: :challenges do
+    resource :discussion, only: :show
```
nemaniarjun: Do we plan to move other controllers to this base controller as well? That seems to be useful and will DRY our code quite a bit :100:

piotrekpasciak: I'll create issue for that. But that's a lot of work and needs to be done in separate PR.

PR #1229, opened 2020-03-11
Display latest Discourse topics on homepage

piotrekpasciak:
Monday: 
https://aicrowd.monday.com/boards/363093650/pulses/476853257


app/views/landing_page/index.html.erb
```diff
@@ -160,14 +160,28 @@
...
-      <%= render partial: 'discourse_top_contributors',
-                 locals: {
-                   discourse_top_contributors_fetch: @discourse_top_contributors_fetch,
-                   discourse_top_contributors:       @discourse_top_contributors
-                 }
-      %>
+      <div class="col-sm-12 col-md-4">
+        <header class="section-header-decorative">
+          <div>
+            <h4>DISCOURSE TOP CONTRIBUTORS</h4>
```
skbly7: No please, we don't it all CAPS here 😅 

And make the heading as just "Top Contributors"

app/views/landing_page/index.html.erb
```diff
@@ -160,14 +160,28 @@
             <h4>DISCOURSE LATEST POSTS</h4>
```
skbly7: "Latest Discussions"

app/views/shared/discourse/_posts_list.html.erb
```diff
@@ -1,37 +1,35 @@
-<div class="sub-section-pb-sm">
-  <% if defined?(discourse_category) %>
-    <header class="section-header">
-      <div>
-        <h4><%= discourse_category %></h4>
-      </div>
-    </header>
-  <% end %>
+<% if defined?(discourse_category) %>
```
skbly7: UI needs to be changed as discussed on Slack.

Similar to https://aicrowd-design.netlify.com/template-challenge-discussion but with users involved in the discussion on right and so on.

app/services/discourse/fetch_latest_posts_service.rb
```diff
@@ -0,0 +1,30 @@
+module Discourse
+  class FetchLatestPostsService < ::Discourse::BaseService
+    def initialize
+      @client = prepare_http_client
+    end
+
+    def call
+      return failure('Discourse API client couldn\'t be properly initialized.') if client.nil?
+      return failure('DISCOURSE_API_LATEST_POSTS_QUERY_ID ENV variable is missing.') if ENV['DISCOURSE_API_LATEST_POSTS_QUERY_ID'].blank?
```
skbly7: 💯 to add it in here already

app/views/landing_page/index.html.erb
piotrekpasciak: This text is converted to cap automatically by  CSS anyway.

skbly7: It doesn't matter, in that case we can fix CSS when needed & changes will be applied in bulk.

Having things in all CAPS even in codebase doesn't look good.

PR #1233, opened 2020-03-12
Add acknowledged and send challenge call to organizer

alkeshrails:
Call for challenge form minor improvements
https://aicrowd.monday.com/boards/363093650/pulses/474943048

### DIFF
--- a/app/admin/challenge_call.rb
+++ b/app/admin/challenge_call.rb
@@ -5,6 +5,13 @@ ActiveAdmin.register ChallengeCall do
     end
   end
 
+  batch_action :acknowledge do |ids|
+    batch_action_collection.find(ids).each do |challenge_call|
+      challenge_call.update(acknowledged: true)
+    end
+    redirect_to admin_challenge_calls_path, notice: "The challenge calls are acknowledged successfully"
+  end
+
   controller do
     def find_resource
       scoped_collection.friendly.find(params[:id])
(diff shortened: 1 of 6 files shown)


app/admin/challenge_call.rb
```diff
@@ -5,6 +5,13 @@
     end
   end
 
+  batch_action :acknowledge do |ids|
+    batch_action_collection.find(ids).each do |challenge_call|
+      challenge_call.update(acknowledged: true)
```
piotrekpasciak: You should probably use `update_all(acknowledged: true, updated_at: Time.current)` to update all records in one request.

Otherwise use `update!` as using `update` is not reliable. We won't know if action worked correctly.

app/admin/challenge_call.rb
```diff
@@ -5,6 +5,13 @@
     end
```
skbly7: Where have you changed the code to send emails to only Organizers and not to all admins?

alkeshrails: @skbly7 Check here https://github.com/AIcrowd/AIcrowd/pull/1233/files#diff-1936ba5d4beb48e633199638a0ffe092R10
also https://github.com/AIcrowd/AIcrowd/pull/1233/files#diff-1936ba5d4beb48e633199638a0ffe092R10
here I set challenge_call id in participant id, is it okay on template?

skbly7: No, the requirement/intention is to sent to challenge_call's organisation's members, not the email of response.

Example:
for: https://www.aicrowd.com/admin/challenge_calls/challenges-on-aicrowd/challenge_call_responses/69

Send email to:
https://www.aicrowd.com/organizers/aicrowd/members

PR #1234, opened 2020-03-12
Update Discussion page with correct layout

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/476851676


app/views/challenges/discussions/show.html.erb
```diff
@@ -1,21 +1,69 @@
...
+                  <div class="avatar-group col-2">
+                    <% discourse_topic['posters'].each_with_index do |poster, index| %>
+                      <%= link_to discourse_user_url(poster['username']) do %>
+                        <%= image_tag poster['avatar_url'], class: "avatar #{'list-group-discussion' if index.zero?}", alt: 'Poster Avatar' %>
+                      <% end %>
+                    <% end %>
+                  </div>
+
+                  <div class="topic-comments">
+                    <img src="https://d33wubrfki0l68.cloudfront.net/5280a0d099efb51753ec327b9c22af89ce52266e/cabdf/assets/img/icon-comment.svg">
```
nemaniarjun: This does not seem correct.
The image is in this repo. It should use asset path instead hard-coded url

piotrekpasciak: I've updated the code.

PR #1237, opened 2020-03-13
Add leaderboard filters

alkeshrails:
image
image

### DIFF
--- a/app/assets/javascripts/controllers/leaderboards_controller.js
+++ b/app/assets/javascripts/controllers/leaderboards_controller.js
@@ -79,33 +79,3 @@ Paloma.controller('Leaderboards', {
   }
 });
 
-$(document).ready(function() {
-  $('#participant-country').change(function() {
-    country_name = $('option:selected',this).val();
-    if(!country_name.trim())
-    {
-      $('#participant-affiliation').find('option').remove().end();
-      $('#participant-affiliation').append($("<option></option>").attr("value",'').text('All affiliations'));
-    }
-    challenge_id = $(this).data('challenge-id');
-    meta_challenge_id = $('#leaderboards-div').data('meta-challenge-id');
-    challenge_round_id = $('#challenge_round_id').val();
-    post_challenge = $('#post_challenge').val();
-
-    var data = {
-      'country_name': country_name,
-      'challenge_round_id': challenge_round_id,
-      'post_challenge': post_challenge
-    }
-
-    if(meta_challenge_id !== '')
-    {
-      data['meta_challenge_id'] = meta_challenge_id
-    }
-
-    $.ajax({
-      url: "/challenges/" + challenge_id + "/leaderboards/get_affiliation.js",
-      data: data
-    });
-  });
-});
--- a/app/assets/javascripts/controllers/leaderboards_controller.js
+++ b/app/assets/javascripts/controllers/leaderboards_controller.js
@@ -78,3 +78,34 @@ Paloma.controller('Leaderboards', {
     pauseAndPlay();
   }
 });
+
+$(document).ready(function() {
+  $('#participant-country').change(function() {
+    country_name = $('option:selected',this).val();
+    if(!country_name.trim())
+    {
+      $('#participant-affiliation').find('option').remove().end();
+      $('#participant-affiliation').append($("<option></option>").attr("value",'').text('All affiliations'));
+    }
+    challenge_id = $(this).data('challenge-id');
+    meta_challenge_id = $('#leaderboards-div').data('meta-challenge-id');
+    challenge_round_id = $('#challenge_round_id').val();
+    post_challenge = $('#post_challenge').val();
+
+    var data = {
+      'country_name': country_name,
+      'challenge_round_id': challenge_round_id,
+      'post_challenge': post_challenge
+    }
+
+    if(meta_challenge_id !== '')
+    {
+      data['meta_challenge_id'] = meta_challenge_id
+    }
+
+    $.ajax({
+      url: "/challenges/" + challenge_id + "/leaderboards/get_affiliation.js",
+      data: data
+    });
+  });
+});
--- a/app/assets/stylesheets/application-redesign.css.scss
+++ b/app/assets/stylesheets/application-redesign.css.scss
@@ -3164,6 +3164,11 @@ input[type="button"].btn-block {
   margin: 0 1.25rem .5rem 1.25rem;
 }
 
+.filter-dropdown-form {
+  display: flex;
+  align-items: center;
+}
+
 .dropdown form hr {
   margin: 1rem 0 .75rem 0 !important;
 }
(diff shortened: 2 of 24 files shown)


app/models/participant.rb
```diff
@@ -249,4 +249,23 @@ def has_accepted_participation_terms?
 
     return true
   end
+
+  def self.countries(p_ids)
```
piotrekpasciak: Class methods should be placed above instance methods:
https://github.com/rubocop-hq/ruby-style-guide#consistent-classes

app/views/leaderboards/index.html.erb
```diff
@@ -13,17 +13,44 @@
...
                             class: round_pills_tab_classes(challenge_round, @current_round) %>
+
               </li>
             <% end %>
           </ul>
           <!-- / Challenge Round Pills  -->
-        <% end %>
+          <!-- country filter -->
+          <div class="row">
+            <%= form_tag challenge_leaderboards_path(challenge_id: @challenge.friendly_id), method: :get, remote: true, class: 'filter-dropdown-form' do %>
```
piotrekpasciak: `form_tag` is a legacy helper. Please use form_with:
https://guides.rubyonrails.org/form_helpers.html

app/controllers/leaderboards_controller.rb
```diff
@@ -59,4 +50,97 @@ def set_current_round
...
+  end
+
+  def affiliation_participant_ids
+    participant_country_leaderboards.pluck(:submitter_id)
+  end
+
+  # get team participant
+  def get_team_participant_ids(team_ids)
+    TeamParticipant.where(team_id: team_ids).pluck(:participant_id)
+  end
```
piotrekpasciak: That's really a lot of logic to put into controller class. Would it be possible to move all this logic to Service (Leaderboards::FilterService) class and write some basic test coverage?

https://github.com/AIcrowd/AIcrowd/pull/1237/files#diff-554837b0b5d7620fd41e48dc0b1d4611R72-R145

app/controllers/leaderboards_controller.rb
```diff
@@ -1,34 +1,21 @@
...
+    if params[:country_cd].present?
+      @leaderboards = Leaderboard.where(id: filter_leaderboard_ids).page(params[:page]).per(10).order(:seq)
     else
-      policy_scope(Leaderboard)
-        .where(challenge_round_id: @current_round&.id.to_i)
-        .page(params[:page])
-        .per(10)
-        .order(:seq)
+      @leaderboards = @leaderboards.page(params[:page]).per(10).order(:seq)
+      @vote             = @challenge.votes.where(participant_id: current_participant.id).first if current_participant.present?
```
piotrekpasciak: You should use `find_by` instead of `where(condition).first` in this line and line below.

app/controllers/leaderboards_controller.rb
```diff
@@ -1,34 +1,21 @@
...
-        .page(params[:page])
-        .per(10)
-        .order(:row_num)
-    elsif @post_challenge
-      policy_scope(OngoingLeaderboard)
-        .where(challenge_round_id: @current_round&.id.to_i)
-        .page(params[:page])
-        .per(10)
-        .order(:seq)
+    @countries = participant_countries if current_participant
```
piotrekpasciak: `current_participant.present?` would improve readability.

alkeshrails: Done

app/controllers/leaderboards_controller.rb
alkeshrails: Done

app/controllers/leaderboards_controller.rb
alkeshrails: Done

app/views/leaderboards/_filter_leaderboards.html.erb
```diff
@@ -0,0 +1,62 @@
...
+      </div>
+
+      <button type="submit" class="btn btn-link btn-sm apply-filter-btn">Filter</button>
+
+    <% end %>
+  </div>
+</div>
+<!-- /country filter -->
+<script type="text/javascript">
+  $("form").submit(function(){
```
piotrekpasciak, line 33: This script should be moved to StimulusJS controller.

app/javascript/stimulus_js/controllers/leaderboards_controller.js
```diff
@@ -0,0 +1,34 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ['participantCountry', 'participantAffiliation']
+
+  filterURL(event){
+    event.preventDefault();
+
+    let url = (window.location.origin+window.location.pathname)
```
piotrekpasciak, line 9: Why do you need to put this code inside of `()`?

Please add spaces between `+`.

app/javascript/stimulus_js/controllers/leaderboards_controller.js
```diff
@@ -0,0 +1,34 @@
...
+
+    let params = ''
+    let country = this.participantCountryTarget.selectedOptions[0].innerText
+    let countryPresent = country != '' && country != 'select-country'
```
piotrekpasciak, line 13: In javascript you should use '!==` and `===` operators. Not checking if variable has both value and correct type leads to a lot of nasty bugs in code.

app/javascript/stimulus_js/controllers/leaderboards_controller.js
```diff
@@ -0,0 +1,34 @@
...
+
+    let affiliation = this.participantAffiliationTarget.selectedOptions[0].innerText
+    let affiliationPresent = affiliation != '' && affiliation != 'select-affiliation'
```
piotrekpasciak, line 16: Like above.

app/assets/javascripts/controllers/leaderboards_controller.js
```diff
@@ -78,3 +78,34 @@ Paloma.controller('Leaderboards', {
     pauseAndPlay();
   }
 });
+
+$(document).ready(function() {
```
piotrekpasciak, line 82: Can we use Stimulus Controller and ES6 instead?  New javascript code should not be written in legacy way.

app/assets/stylesheets/application-redesign.css.scss
```diff
@@ -14042,6 +14047,62 @@ html {
...
+  border-top: 1px solid #D9DCDC;
+  padding: 1rem 0 1rem 1.5rem;
+}
+
+.leaderboard-filters {
+  font-size: 14px;
+  padding-bottom: 1rem;
+}
+
+@media (max-width: 767.98px) {
```
piotrekpasciak, line 14063: It's more clean to use media query inside fo class that's affected by it. Take a look on this example:
https://github.com/AIcrowd/AIcrowd/blob/master/app/assets/stylesheets/participants/activity_heatmap.scss#L1-L8

app/assets/stylesheets/application-redesign.css.scss
```diff
@@ -14042,6 +14047,62 @@ html {
...
+  align-items: flex-start;
+  display: flex;
+  flex: 1 0 auto;
+}
+
+.leaderboard-filters form .form-group {
+  margin: 0 .5rem 0 0;
+}
+
+@media (max-width: 767.98px) {
```
piotrekpasciak, line 14079: Like above.

app/assets/stylesheets/application-redesign.css.scss
```diff
@@ -14042,6 +14047,62 @@ html {
...
+    margin: 0 .5rem .75rem 0;
+  }
+}
+
+.leaderboard-filters form select {
+  font-size: 14px;
+  width: 160px;
+}
+
+@media (max-width: 575.98px) {
```
piotrekpasciak, line 14091: Like above.

app/models/participant.rb
```diff
@@ -141,6 +141,26 @@ def reconfirmation_required?
     false
   end
 
+  def self.countries(p_ids)
```
piotrekpasciak, line 144: Those 4 method don't seem to belong here. It would be better to create a separate class like Country under app/lib directory.

app/policies/leaderboard_policy.rb
```diff
@@ -49,6 +45,10 @@ def participant_sql(participant)
             )
           )
       SQL
+      else
+        participant_id = 0
+        "leaderboards.submitter_type = 'Participant' AND leaderboards.submitter_id = #{participant_id} OR leaderboards.challenge_id IN (SELECT c.id FROM challenges c WHERE c.show_leaderboard IS TRUE AND c.private_challenge IS FALSE UNION SELECT c.id FROM challenges c WHERE c.show_leaderboard IS TRUE  AND c.private_challenge IS TRUE)"
```
piotrekpasciak, line 50: Please refactor it to ActiveRecord or move under `app/queries` directory to separate class.  We should avoid plain SQL queries as they are difficult to maintain.

app/policies/ongoing_leaderboard_policy.rb
```diff
@@ -12,32 +12,31 @@ def participant_sql(participant)
...
+                WHERE invitations.challenge_id = c.id
+                AND invitations.email = '#{email}'
+              )
+            )
+        SQL
       else
-        participant_id       = 0
-        email                = nil
-        participant_team_ids = nil
+        participant_id = 0
```
piotrekpasciak, line 37: Like above.

PR #1238, opened 2020-03-13
Install dalli Memcache client

piotrekpasciak:
Documentation:
https://devcenter.heroku.com/articles/building-a-rails-3-application-with-memcache

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -14,9 +14,6 @@ Rails.application.configure do
   config.i18n.fallbacks                    = true
   config.log_formatter                     = ::Logger::Formatter.new
 
-  # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
-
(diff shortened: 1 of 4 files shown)


config/environments/production.rb
```diff
@@ -48,5 +45,15 @@
     user_name: ENV["SMTP_USERNAME"]
   }
   config.action_mailer.default_url_options = { host: ENV["SMTP_DOMAIN"] }
+  config.cache_store = :mem_cache_store,
+                       (ENV['MEMCACHEDCLOUD_SERVERS'].to_s).split(','),
```
alkeshrails, line 49: Don't get the logic to split env var.

piotrekpasciak: https://devcenter.heroku.com/articles/building-a-rails-3-application-with-memcache

Documentation has this config. example to allow multiple servers when needed.

PR #1239, opened 2020-03-15
Bugfix: Development environment breaking when discourse not setted up

skbly7:
Removing the dependency of having Discourse set up, some views are breaking otherwise


skbly7:
Minor change, merging without peer approval so other dev branches can use updated master

PR #1241, opened 2020-03-16
Challenge Organizer multi-select2 box

nemaniarjun:


app/views/challenges/update.js.erb
```diff
@@ -1,33 +1,31 @@
...
-
-    if (updateSuccessful === 'true') {
-      flashMessagesWrapper.innerHTML = "<%= j render 'layouts/flash_message', type: 'notice', message: 'Challenge updated.' %>";
-    } else {
-      flashMessagesWrapper.innerHTML = "<%= j render 'layouts/flash_message', type: 'error', message: 'This tab couldn\'t be saved because of errors. Please correct them.' %>";
-    }
-  }, window.pageYOffset / 4);
-
-
+    const challengeForWrapper  = document.getElementById('challenges-form__wrapper');
```
piotrekpasciak: Is there any particular reason why you use 4 spaces for indentation for javascript? I think we should use 2 spaces everywhere. Can you update  your editor settings?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_03_09_134625) do
+ActiveRecord::Schema.define(version: 2020_03_05_150815) do
```
piotrekpasciak: I don't see any migrations in your PR, please remove the changes from schema.rb file.

PR #1242, opened 2020-03-17
Assign participant object to response from Discourse

piotrekpasciak:


app/views/challenges/discussions/show.html.erb
```diff
@@ -32,14 +32,14 @@
 
                     <div>
                       <%= link_to discourse_topic['title'], discourse_topic_url(discourse_topic['slug']), class: 'topic-title text-link' %>
-                      <span>By <%= link_to author['username'], discourse_user_url(author['username']), class: 'text-link' %> <%= discourse_time_ago(discourse_topic['created_at']) %></span>
+                      <span>By <%= link_to author['participant']&.name, discourse_user_url(author['participant']&.name), class: 'text-link' %> <%= discourse_time_ago(discourse_topic['created_at']) %></span>
```
alkeshrails: As you added try(&) here, if author['participant']&.name so what will be the link here.

app/views/challenges/discussions/show.html.erb
```diff
@@ -32,14 +32,14 @@
...
                     </div>
                   </div>
 
                   <div class="avatar-group col-2">
                     <% discourse_topic['posters'].each_with_index do |poster, index| %>
-                      <%= link_to discourse_user_url(poster['username']) do %>
-                        <%= image_tag poster['avatar_url'], class: "avatar #{'list-group-discussion' if index.zero?}", alt: 'Poster Avatar' %>
+                      <%= link_to discourse_user_url(poster['participant']&.name) do %>
```
alkeshrails: same as above.

app/views/landing_page/_discourse_top_contributors.html.erb
```diff
@@ -4,9 +4,9 @@
     <% discourse_top_contributors.each do |top_contributor| %>
       <!-- user -->
       <li class="list-group-item">
-        <%= link_to discourse_user_url(top_contributor['username']) do %>
-          <%= image_tag top_contributor['avatar_url'], class: "avatar" %>
-          <span class="text-link"><%= top_contributor['username'] %></span>
+        <%= link_to discourse_user_url(top_contributor['participant']&.name) do %>
```
alkeshrails: same as above.

app/views/shared/discourse/_topics_list.html.erb
```diff
@@ -16,14 +16,14 @@
 
         <div>
           <%= link_to discourse_topic['title'], discourse_topic_url(discourse_topic['slug']), class: 'topic-title text-link' %>
-          <span>By <%= link_to author['username'], discourse_user_url(author['username']), class: 'text-link' %> <%= discourse_time_ago(discourse_topic['created_at']) %></span>
+          <span>By <%= link_to author['participant']&.name, discourse_user_url(author['participant']&.name), class: 'text-link' %> <%= discourse_time_ago(discourse_topic['created_at']) %></span>
```
alkeshrails: same as above.

spec/services/discourse/fetch_top_contributors_service_spec.rb
```diff
@@ -28,7 +28,6 @@
         expect(response.size).to eq 5
         expect(response.first['username']).to eq 'kelleni2'
         expect(response.first['score']).to eq 751
-        expect(response.first['avatar_url']).to eq 'http://localhost:3000/assets/users/user-avatar-default.svg'
```
alkeshrails: we can test new path of default URL here.

app/views/challenges/discussions/show.html.erb
piotrekpasciak: If we don't have Participant from Discourse in our database, link to `#` will be displayed doesn't take user anywhere. Not sure how to handle this scenario better.

spec/services/discourse/fetch_top_contributors_service_spec.rb
piotrekpasciak: Alright. I'll update the spec file.

app/views/challenges/discussions/show.html.erb
alkeshrails: we can add `if` condition before span to check `author['participant']&.name`

piotrekpasciak: I've added if statements everywhere.

spec/services/discourse/fetch_top_contributors_service_spec.rb
```diff
@@ -28,7 +30,7 @@
...
+        expect(response.first['participant']).to eq participant
```
alkeshrails: I think, it would be good to test  the default image url case as well

PR #1250, opened 2020-03-18
Fix delete member from organiser

alkeshrails:


app/controllers/members_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
 
   def destroy
     participant = Participant.friendly.find(params[:id])
-    participant.update(organizer_id: nil)
+    @organizer.participants.destroy(participant)
```
piotrekpasciak: Let's follow policy of failing early. Please use `destroy!` instead of `destroy`.

PR #1260, opened 2020-03-19
Create category tag to challenge form

alkeshrails:
Create new category to challenge/edit
https://aicrowd.monday.com/boards/363093650/pulses/499058864

### DIFF
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -3,6 +3,10 @@ $(document).on('click', '.challenge-invite-participant', function(){
   $('.challenge-invite-participant').attr("disabled", true);
 });
 
+$(document).ready(function() {
+  $('.category_select2').select2({
+    tags: true
+  });
 
 $(document).ready(function() {
   $('input[name="status"]:radio').change(function(){
--- a/app/assets/javascripts/controllers/challenges_controller.js
+++ b/app/assets/javascripts/controllers/challenges_controller.js
@@ -4,11 +4,10 @@ $(document).on('click', '.challenge-invite-participant', function(){
 });
 
 $(document).ready(function() {
+  $('.select2').select2();
   $('.category_select2').select2({
     tags: true
   });
-
-$(document).ready(function() {
   $('input[name="status"]:radio').change(function(){
     $('#challenge-statuses').addClass('filter-active');
   });
@@ -73,6 +72,9 @@ Paloma.controller('Challenges', {
     edit: function () {
         $(document).ready(function() {
             $('.select2').select2();
+            $('.category_select2').select2({
+              tags: true
+            });
         });
         const currentUrl  = new URL(window.location);
         const currentStep = currentUrl.searchParams.get('step');
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -9,7 +9,6 @@ class ChallengesController < ApplicationController
   before_action :set_challenge_rounds, only: [:edit, :update]
   before_action :set_filters, only: [:index]
   before_action :set_organizers_for_select, only: [:new, :create, :edit, :update]
-  before_action :set_categories_for_select, only: [:new, :create, :edit, :update]
 
   respond_to :html, :js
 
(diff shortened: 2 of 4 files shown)


app/helpers/challenges_helper.rb
```diff
@@ -75,6 +75,10 @@ def check_selected_prize(prize, params)
     params[:prize].dig(:prize_type).include?(prize.to_s)
   end
 
+  def select_option(challenge)
```
piotrekpasciak: This is very generic name for method. Please use more specific name like `categories_select_options`.

app/controllers/challenges_controller.rb
```diff
@@ -199,15 +197,12 @@ def create_invitations
 
   def create_challenge_category
     @challenge.category_challenges.destroy_all if @challenge.category_challenges.present?
-    params[:challenge][:category_ids].each do |category_id|
-      @challenge.category_challenges.create(category_id: category_id)
+    params[:challenge][:category_names].reject(&:empty?).each do |category_name|
+      category = Category.find_or_create_by!(name: category_name)
+      @challenge.category_challenges.create(category_id: category.id)
```
piotrekpasciak: Please always use methods with `!` when you don't handle validations in code. `create!` instead of `create` here.

app/views/challenges/tabs/_details_form.html.erb
```diff
@@ -16,7 +16,7 @@
 
   <div class="form-group">
     <span class="label-text">Category</span>
-    <%= f.select :category_ids, @categories_for_select, {}, {class: "select2 form-control", multiple: true} %>
+    <%= f.select :category_names, categories_select_options(@challenge), {}, {class: "select2 category_select2 form-control", multiple: true} %>
```
nemaniarjun: no need to add the class "category_select" I renamed it to just "select2". See other comment for my doubt regarding this.

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -3,8 +3,11 @@ $(document).on('click', '.challenge-invite-participant', function(){
   $('.challenge-invite-participant').attr("disabled", true);
 });
 
-
 $(document).ready(function() {
+  $('.select2').select2();
```
nemaniarjun: This is not needed here. We only need to fire these event on the challenge edit page. Paloma does this for us in the edit function below.

app/assets/javascripts/controllers/challenges_controller.js
```diff
@@ -69,6 +72,9 @@ Paloma.controller('Challenges', {
     edit: function () {
         $(document).ready(function() {
             $('.select2').select2();
```
nemaniarjun:
```
 $('.select2').select2({ tags: true });
```

is just this one line not enough? 
will adding "Tags: true" to the other select2 (organizers dropdown)  cause issues?

app/controllers/challenges_controller.rb
```diff
@@ -63,7 +63,7 @@ def create
 
     if @challenge.save
       update_challenges_organizers if params[:challenge][:organizer_ids].present?
-      update_challenge_categories if params[:challenge][:category_ids].present?
+      create_challenge_category if params[:challenge][:category_names].present?
```
nemaniarjun: Let the name be ```update_challenge_categories``` as it is much more clear

app/helpers/challenges_helper.rb
```diff
@@ -75,6 +75,10 @@ def check_selected_prize(prize, params)
     params[:prize].dig(:prize_type).include?(prize.to_s)
   end
 
+  def categories_select_options(challenge)
+    options_from_collection_for_select(Category.all, :name, :name, { selected: challenge.categories.pluck(:name) })
```
nemaniarjun: This is an interesting way of doing this. 
do we need the function `set_categories_for_select` in challenges_controller when we are doing this here?

app/controllers/challenges_controller.rb

app/helpers/challenges_helper.rb
alkeshrails: Thanks

app/assets/javascripts/controllers/challenges_controller.js
alkeshrails: If I remove this it will cause issue on your Organizers select2 on [CreateChallenge](https://tinyurl.com/varfaan)

app/assets/javascripts/controllers/challenges_controller.js
alkeshrails: I removed this and add new class, because in my case I need `tags : true`

app/views/challenges/tabs/_details_form.html.erb
alkeshrails: In my case I need `tags: true` so instead of add tag on select2 I add create a new class `category_select2` and add tags there.

Issue #1262, opened 2020-03-20
Organize the seeding scripts


spMohanty:
Its chaos in the current seeding script.
We should organize it a bit more, and also include meaningful data there.

PR #1263, opened 2020-03-20
Addresses #1262 - Fix some broken seeds which were causing validation…

spMohanty:
… errors


skbly7:
Sorry, you will need to rebase this PR, the sidekick 6.0.15 which is being referred in your PR is no longer available. 

Let us know in case you want to offload further steps to @Rails team, we can take the PR forward accordingly.

Issue #1264, opened 2020-03-20
nil participant error when rendering tooltips


spMohanty:
The master branch currently gives the following error : 

```
undefined method `id' for nil:NilClass
```

```
<% 
    uniq_id = rand(1000)
    random_class = 'tooltip-user-wrapper-' + uniq_id.to_s + '-' + participant.id.to_s
%>
<div class="tooltip-user-wrapper <%= random_class %>">
    <div style="display: table-row;" data-container=".<%= random_class %>" data-html="true" data-toggle="tooltip" 

```

<img width="956" alt="Screenshot 2020-03-20 at 16 20 26" src="https://user-images.githubusercontent.com/1581312/77178362-43fe5900-6ac7-11ea-9d4d-5caeaa0bfd7e.png">

spMohanty:
@piotrekpasciak : This is hapening because [here](https://github.com/AIcrowd/AIcrowd/blob/master/app/views/shared/discourse/_topics_list.html.erb#L29) we are not checking if the participant is actually present.

This is caused by the divergence between the list of participants on the rails-web-app and the discourse. Unlikely to happen on production, but will probably happen a lot during development.

skbly7:
Fixed after #1264, #1267

PR #1267, opened 2020-03-20
Addresses #1264 - Hotfix for rendering nil users for discourse posts

spMohanty:


app/views/participants/_tooltip_profile.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="tooltip-inner-top">
+x<div class="tooltip-inner-top">
```
nemaniarjun: accidental change?

db/schema.rb
```diff
@@ -783,6 +783,7 @@
     t.float "temporary_variation"
     t.integer "ranking", default: -1, null: false
     t.integer "ranking_change", default: 0, null: false
+    t.boolean "agreed_to_organizers_newsletter", default: true, null: false
```
nemaniarjun: Unnecessary change. 
Remove this file from the commit

app/views/shared/discourse/_topics_list.html.erb
```diff
@@ -26,7 +26,11 @@
 
       <div class="avatar-group col-2">
         <% discourse_topic['posters'].take(4).each_with_index do |poster, index| %>
-          <%= render partial: 'participants/user_link', locals: { participant: poster['participant'], username: false } %>
+          <% if poster['participant'].present? %>
```
nemaniarjun: This change should be made in the partial as there might be other places where the partial is used

app/views/participants/_tooltip_profile.html.erb
spMohanty: Yes, an accidental change. Fixed in a subsequent commit.

db/schema.rb
spMohanty: agree, this also got added by mistake.
But this version of `schema.rb` should anyway be reflected in the master branch right ?

app/views/shared/discourse/_topics_list.html.erb
spMohanty: Agree. 
I will let @skbly7 deal with the refactoring there, as he set that whole thing up.
At the moment, pushing this to ensure the dev setup is not broken atleast. Others had also faced the same issue.

db/schema.rb
piotrekpasciak: This change actually is good. Master branch is missing, because I incorrectly resolved the conflics. I recommend keeping it in this PR. Otherwise I'll have to readd it in other PR.

app/views/shared/discourse/_topics_list.html.erb
skbly7: ACK, no harm and this should anyway never happen in the production environment. 

As we discussed previously, we need to add necessary checks in Discourse integration so it doesn’t active in the dev environment, which still seems to be the case. @piotrekpasciak

PR #1270, opened 2020-03-20
1269 spm fix rendering of challenge categories

spMohanty:
This is a hot fix of sorts. 
And this has to be properly addressed by @jsnrynlds  in the updated design.

Also, this pull request has a dependency on this one : https://github.com/AIcrowd/AIcrowd/pull/1267


db/schema.rb
```diff
@@ -783,6 +783,7 @@
     t.float "temporary_variation"
     t.integer "ranking", default: -1, null: false
     t.integer "ranking_change", default: 0, null: false
+    t.boolean "agreed_to_organizers_newsletter", default: true, null: false
```
skbly7: This file needs to be removed from this PR.

PR #1274, opened 2020-03-20
converting sql query to an efficient one

nikhilrayaprolu:


app/controllers/participants_controller.rb
```diff
@@ -22,8 +22,7 @@ def show
                              end
     @discourse_posts = @discourse_posts_fetch.value
 
-    @categories = @participant.challenges.joins(:categories).pluck('categories.name')
-                    .inject(Hash.new(0)) { |h, e| h[e] += 1 ; h }
+    @categories = @participant.challenges.joins(:categories).group('categories.name').reorder('catogories.name').count
```
piotrekpasciak: `.reorder('catogories.name')` there is a typo here.

app/controllers/participants_controller.rb
```diff
@@ -22,8 +22,7 @@ def show
                              end
     @discourse_posts = @discourse_posts_fetch.value
 
-    @categories = @participant.challenges.joins(:categories).pluck('categories.name')
-                    .inject(Hash.new(0)) { |h, e| h[e] += 1 ; h }
+    @categories = @participant.challenges.joins(:categories).group('categories.name').reorder('categories.name').count
     if @categories.count == 0
```
piotrekpasciak: Are you sure this line will work, when you already called `.count` in line above?

nikhilrayaprolu: yes, above query provides a hash
and the below count is a count on the number of elements in the hash

PR #1277, opened 2020-03-21
decay and mu-3 sigma implementation for rating

nikhilrayaprolu:


app/controllers/user_rating_controller.rb
```diff
@@ -23,7 +21,7 @@ def post_new_participant_ratings
     end
     user_rating_service = UserRatingService.new(params[:round_id])
     teams_participant_ids, new_team_ratings, new_team_variations = params[:participant_ids], params[:final_rating], params[:final_variation]
-    if new_team_ratings.blank? && new_team_variations.blank?
```
spMohanty: Are the terms `ratings` and `variations` referring to `mu` and `sigma` ?
If so, it might make sense to use a consistent naming convention here too.

app/services/rating_api_service.rb
```diff
@@ -10,7 +10,7 @@ def call(challenge_rounds)
     }.to_json
     headers = { 'Content-Type' => 'application/json' }
     begin
-      result = self.class.post('/startroundcalculations/', body: body, headers: headers)
+      result = self.class.post('http://localhost:8888/startroundcalculations/', body: body, headers: headers)
```
spMohanty: The end point location should ideally be picked up from an environment variable. 
cc. @skbly7

app/queries/participant_rating_ranks_query.rb
```diff
@@ -6,16 +6,25 @@ def initialize(relation = Participant.all)
   end
 
   def call()
-    ranks = @relation.where('rating is not null').select('id, rank() over (order by rating desc) as current_ranking, ranking as previous_ranking, ranking_change').reorder('ranking asc')
+    @relation.where('fixed_rating is not null').map do |participant|
+      user_final_rating = UserRating.where("participant_id=#{participant.id} and rating is not null").reorder('created_at desc').first
+      time_difference = (Time.current.to_date - user_final_rating[:created_at].to_date)
```
spMohanty: Will the value of `user_final_rating[:created_at]` be updated as soon as we update the scores here again because of the decay ?

Ideally, the time-difference here should refer to the number of days since the last reputation update (by anything other than a decay).
For example, if the last competition that a user took part in, was 6 months ago. In the meantime the reputation has been decaying anyway. And to find the decay associated with today, we take the time_difference as `180days`..

app/services/rating_api_service.rb
skbly7: We already have environment variable for the same.

app/queries/participant_rating_ranks_query.rb
```diff
@@ -6,16 +6,25 @@ def initialize(relation = Participant.all)
   end
 
   def call()
-    ranks = @relation.where('rating is not null').select('id, rank() over (order by rating desc) as current_ranking, ranking as previous_ranking, ranking_change').reorder('ranking asc')
+    @relation.where('fixed_rating is not null').map do |participant|
+      user_final_rating = UserRating.where("participant_id=#{participant.id} and rating is not null").reorder('created_at desc').first
```
piotrekpasciak: Please use prepared statements here instead of string interpolation.

app/queries/participant_rating_ranks_query.rb
```diff
@@ -6,16 +6,25 @@ def initialize(relation = Participant.all)
...
+      time_difference = (Time.current.to_date - user_final_rating[:created_at].to_date)
+      time_difference = time_difference.to_i
+      total_number_of_days = 365
+      updated_rating = participant.fixed_rating * (Math.exp(-time_difference/total_number_of_days))
+      participant.update({rating: updated_rating})
```
piotrekpasciak: You should use `update!` method to not fail silently.

app/queries/participant_rating_ranks_query.rb
```diff
@@ -6,16 +6,25 @@ def initialize(relation = Participant.all)
...
+      UserRating.create(participant_id: participant.id, rating: updated_rating, variation: user_final_rating['variation'], challenge_round: user_final_rating['challenge_round'])
```
piotrekpasciak: You should use `create!` method to make user UserRating was actually created.

app/queries/participant_rating_ranks_query.rb
piotrekpasciak: Using `find_by` instead of `where(condition).first` is recommended approach in Ruby community.

nikhilrayaprolu: `participant_id=#{participant.id} and rating is not null and challenge_round is not null` how can not null be also return in find_by? @piotrekpasciak

skbly7: @nikhilrayaprolu Irrespective of `find_by` can return above thing (your question). 
In case you are using `where(condition).first` for some specific reason i.e. scenario described by you, I would like that to be added as part of our codebase, even if only as comment.

So in future it doesn't end up being magic -- which no one understand how it is working.

app/queries/participant_rating_ranks_query.rb
nikhilrayaprolu: Yes, it works in the way you think. the user_final_rating is being brought from user history, a new row created every time the rating gets updated.

The user_ratings that are changed every day are created with no challenge id, so I query the recent one that has a challenge_id -- that provides me the `created_at`.

Here the assumption is the user_rating is created the night the challenge has ended - so both challenge end date and the created_at of the user rating would differ by 1 day. 

Since this could be a silly assumption, I am modifying this by making a join with the challenge_round and then finding the exact challenge end date.

app/controllers/user_rating_controller.rb
nikhilrayaprolu: modified all mu and sigma to rating and variation.

piotrekpasciak:
@nikhilrayaprolu I don't see schema.rb changes from migration you have added in this PR.

piotrekpasciak:
@nikhilrayaprolu I don't see schema.rb changes from migration you have added in this PR.

app/queries/participant_rating_ranks_query.rb
nikhilrayaprolu: Resolved accordingly

nikhilrayaprolu:
@skbly7 resolved all the issues,
image


Also implemented challenge tooltip
image

app/services/user_rating_service.rb
```diff
@@ -13,38 +13,42 @@ def leaderboard_query()
...
           new_participant_variations << {"variation" => new_team_variations[i][j]}
         end
       end
     end
     return participant_ids, new_participant_ratings, new_participant_variations
   end
-
+  def update_challenge_permanent
+    unless @temporary
+      @round.update(calculated_permanent: true)
```
piotrekpasciak: Please use `update!` to ensure this line worked.

app/views/participants/_show_reputation.erb
```diff
@@ -4,11 +4,11 @@
     ###
   </span>
 <% else %>
-  <span class="reputation-score" data-toggle="tooltip" title="<%= participant.rating.to_i %> ± <%= participant.variation.to_i %>">
+  <span class="reputation-score" data-toggle="tooltip" title="<%= participant.rating.to_i - 3*participant.variation.to_i %> ± <%= participant.variation.to_i %>">
```
piotrekpasciak: `participant.rating.to_i - 3*participant.variation.to_i` this kind of expressions should be moved to helper method, especially when you use them multiple times like here.

app/views/participants/_masthead.html.erb
```diff
@@ -11,6 +11,7 @@
               <h1>
                 <%= participant.name %>
                 <%= render :partial => 'participants/show_reputation', locals: { participant: participant } %>
+                <%= render :partial => 'participants/show_temporary_reputation', locals: { participant: participant } %>
```
piotrekpasciak: Please use new hash syntax:
https://github.com/rubocop-hq/ruby-style-guide#hash-literals

app/views/participants/_show_reputation.erb
```diff
@@ -4,11 +4,11 @@
...
     <% if defined?(full) and full %>
```
piotrekpasciak: Use `&&` instead of `and` operator.

app/views/participants/_show_temporary_reputation.erb
```diff
@@ -0,0 +1,15 @@
+<span class="reputation-info" style="<%= style ||= '' %>">
+<% if participant.admin %>
+  <span class="reputation-score" data-toggle="tooltip" title="Admin">
+    ###
+  </span>
+<% else %>
+  <span class="reputation-score" data-toggle="tooltip" title="<%= participant.temporary_rating.to_i - 3*participant.temporary_variation.to_i %> ± <%= participant.temporary_variation.to_i %>">
+    <% if defined?(full) and full %>
```
piotrekpasciak: Use && instead of `and`.

app/views/participants/_show_temporary_reputation.erb
```diff
@@ -0,0 +1,15 @@
+<span class="reputation-info" style="<%= style ||= '' %>">
+<% if participant.admin %>
+  <span class="reputation-score" data-toggle="tooltip" title="Admin">
+    ###
+  </span>
+<% else %>
+  <span class="reputation-score" data-toggle="tooltip" title="<%= participant.temporary_rating.to_i - 3*participant.temporary_variation.to_i %> ± <%= participant.temporary_variation.to_i %>">
```
piotrekpasciak: `participant.temporary_rating.to_i - 3*participant.temporary_variation.to_i %> ± <%= participant.temporary_variation.to_i` - Move these expressions to helper.

app/views/participants/_tooltip_profile.html.erb
```diff
@@ -5,6 +5,7 @@
         <br>
         <div class="reputation-info">
             <%= render :partial => 'participants/show_reputation', locals: { participant: participant, style: 'top: 0px;', full: true } %>
+            <%= render :partial => 'participants/show_temporary_reputation', locals: { participant: participant, style: 'top: 0px;', full: true } %>
```
piotrekpasciak: 1. New hash syntax. 
2. Avoid using `style` attribute. You should use css class instead.

app/views/participants/show.html.erb
```diff
@@ -106,3 +106,45 @@
     <!-- /section -->
   </div>
 </div>
+<script>
+    function indexOfArray(val, array) {
```
piotrekpasciak: Move this code to Stimulus controller:
- https://stimulusjs.org/reference/controllers

Example controller:
https://github.com/AIcrowd/AIcrowd/blob/master/app/javascript/stimulus_js/controllers/chart_controller.js

Example usage:
https://github.com/AIcrowd/AIcrowd/blob/master/app/views/insights/_charts/_top_score_vs_time.html.erb#L1-L7

app/views/participants/show.html.erb
```diff
@@ -106,3 +106,45 @@
...
+            }
+        }
+    };
+    var chart = Chartkick.charts["line-chart"];
+    var newOptions = {
+        library: {
+            tooltips: {
+                callbacks: {
+                    label: function(tooltipItem, data) {
+                        console.log("label", tooltipItem, data);
```
piotrekpasciak: Don't leave debuggers like `console.log` in codebase.

app/views/participants/show.html.erb
```diff
@@ -106,3 +106,45 @@
...
+
+                    }
+                }
+            }
+        }
+    };
+    var mergedOptions = jQuery.extend(chart.options,newOptions);
+    chart.setOptions(mergedOptions);
+    pointarray = new Array(chart.chart.data.datasets[0].data.length).fill(0)
+    console.log(pointarray)
```
piotrekpasciak: Like above.

app/views/participants/show.html.erb
```diff
@@ -106,3 +106,45 @@
...
+    data_array = chart.chart.data.datasets[0].data
+    for(j in data_array) {
+        challenge_round_name = chart.dataSource[indexOfArray(data_array[j], chart.dataSource)][2]
+        if(challenge_round_name != '') {
+            console.log(j)
```
piotrekpasciak: like above

app/views/participants/show.html.erb
```diff
@@ -106,3 +106,45 @@
...
+            pointarray[j] = 4
+            console.log(pointarray)
```
piotrekpasciak: Like above.

app/views/rating_leaderboard/_rating_leaderboard.html.erb
```diff
@@ -20,6 +20,7 @@
   </td>
   <td class="<%= cls ||= '' %>">
     <%= render partial: 'participants/show_reputation', locals: { participant: participant, style: 'top: 0px;' } %>
+    <%= render partial: 'participants/show_temporary_reputation', locals: { participant: participant, style: 'top: 0px;' } %>
```
piotrekpasciak: Use CSS class instead of style attribute.

PR #1281, opened 2020-03-24
Fix Review Apps Config

piotrekpasciak:
This fix consists of:
- reusing staging environment config instead of test one (test one should be used only for running specs locally or in Continuous Integration)
- handling multiple domains issue for Review Apps
- setting proper default ENV variables for Review Apps 

Github OAuth will redirect_to staging website for now, because we don't know way to handle multiple domains issue of Review Apps.


app.json
```diff
@@ -1,9 +1,7 @@
 {
   "addons": [
-    "logdna",
     "heroku-postgresql",
-    "redistogo",
```
nemaniarjun: Are we sure this will only affect the review apps?

piotrekpasciak: Yes, because according to Heroku documentation this file is used only by Review apps:
https://devcenter.heroku.com/articles/github-integration-review-apps#setup

config/application.yml.example
```diff
@@ -28,51 +28,3 @@ development: &development
 test:
   <<: *development
   MANDRILL_KEY: "G_rHM-BQ9DF6IG2wKa3-Pw"
-
```
skbly7: Why? Shouldn't we keep staging and production example config?

piotrekpasciak: Because it's not used, this dead and outdated piece of code. We use ENV's from application.yml only in development and test environments:
https://github.com/AIcrowd/AIcrowd/blob/master/config/environments/test.rb#L5

Staging and production doesn't use Figaro.

config/environments/staging.rb
```diff
@@ -1,7 +1,10 @@
 Rails.application.configure do
+  # Handle dynamic domain names of review_apps
+  ENV['DOMAIN_NAME'] = "https://#{ENV['HEROKU_APP_NAME']}.herokuapp.com" if ENV['IS_REVIEW_APP'] == 'true'
+
   config.cache_classes                             = true
   config.eager_load                                = true
-  config.consider_all_requests_local               = false
+  config.consider_all_requests_local               = true
```
skbly7: This looks weird. It should be `false` for staging?

piotrekpasciak: I wanted to show errors traceback. It will help us fix bugs faster on review_apps and staging. Do you prefer to keep hidden errors on  staging?

PR #1282, opened 2020-03-24
Improve og meta tags for submissions

alkeshrails:


app/helpers/meta_tags_helper.rb
```diff
@@ -12,15 +12,25 @@ def meta_title
   end
 
   def meta_description
+    return @submission.challenge.tagline if submissions?
```
nemaniarjun: Not sure why we need this ?

alkeshrails: It is the case, when we are on submissions show page then the description will be given by submission's challenge tagline, code is not yet completed, I'm working on this.

nemaniarjun:
Looks like this contains code also in https://github.com/AIcrowd/AIcrowd/pull/1284 if yes, you can close this PR.

PR #1283, opened 2020-03-24
Bugfix: Fixed wrong logic to check for full teams

skbly7:
Bug reported by @jyotishp and email from participant.

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -43,9 +43,7 @@ class Team < ApplicationRecord
   end
 
   def full?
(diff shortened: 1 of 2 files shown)


app/models/team.rb
```diff
@@ -43,9 +43,7 @@ def invitations_left_clamped
   end
 
   def full?
```
skbly7: Crux: The `full?` was being used to solve two places with different intentions. One being if the team is full with members and one to check if the team has any invitations left.

PR #1285, opened 2020-03-25
Add Organizer::NewsletterEmailMailer and queue it after admin approve

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/497868345

Documentation:
https://mandrillapp.com/api/docs/messages.JSON.html#method=send

### DIFF
--- a/app/admin/newsletter_email.rb
+++ b/app/admin/newsletter_email.rb
@@ -1,14 +1,19 @@
 ActiveAdmin.register NewsletterEmail do
-  actions :index, :show, :new, :create
-  permit_params :bcc, :cc, :subject, :message
+  config.sort_order = :pending_desc
+
+  actions :index, :show
 
   index do
     column :id
+    column :to do |resource|
+      resource.participant&.email
+    end
     column :bcc
     column :cc
     column :subject
     column :message
     column :pending
+    column :participant
     column :created_at
     column :updated_at
     actions default: true do |resource|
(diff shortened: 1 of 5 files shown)


app/admin/newsletter_email.rb
```diff
@@ -19,18 +24,8 @@
...
-      f.input :cc, as: :string
-      f.input :subject, as: :string
-      f.input :message
-    end
-    f.actions
-  end
-
   member_action :approve, method: :post do
-    # TODO: Implement business logic in next Pull Request
+    Organizers::NewsletterEmailJob.perform_now(resource.id)
```
nemaniarjun: Should this be a foreground call ? If we have any errors this would timeout

app/mailers/organizers/newsletter_email_mailer.rb
```diff
@@ -0,0 +1,30 @@
...
+    private
+
+    def format_options(newsletter_email)
+      options = {
+        participant_id:    newsletter_email.participant_id,
+        subject:           newsletter_email.subject,
+        to:                newsletter_email.participant.email,
+        cc:                newsletter_email.cc,
+        bcc:               newsletter_email.bcc,
+        template:          'AICrowd Newsletter Email Template',
```
nemaniarjun:
```suggestion
        template:          'AIcrowd Newsletter Email Template',
```

app/admin/newsletter_email.rb
piotrekpasciak: It should be background call. Good catch.

PR #1290, opened 2020-03-27
Api to check user is login

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/497675888


alkeshrails:
Will create new PR for this.

PR #1291, opened 2020-03-27
Api to check user login

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/497675888


app/controllers/api/session_helper_controller.rb
```diff
@@ -0,0 +1,17 @@
+require 'cgi'
+require 'active_support'
+
+class Api::SessionHelperController < Api::BaseController
+  respond_to :json
+
+  def check_login
+    result      = SessionService.new(params[:aicrowd_cookie]).call
+    participant = Participant.find_by(id: result["warden.user.participant.key"][0]) if result.present? && !result["warden.user.participant.key"].nil?
```
piotrekpasciak: `result.present?` - you could reuse `success?` `failure?` pattern here instead.

Example:
https://github.com/AIcrowd/AIcrowd/blob/master/app/services/challenges/import_service.rb#L16
https://github.com/AIcrowd/AIcrowd/blob/eff2a2fae5e195fea038a078a4528b7c557b7b6c/app/controllers/challenges_controller.rb#L133

app/services/session_service.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def call
+    key_generator = ActiveSupport::KeyGenerator.new(@secret_key_base, iterations: 1000)
+    key_len       = ActiveSupport::MessageEncryptor.key_len(@encrypted_cookie_cipher)
+    secret        = key_generator.generate_key(@salt, key_len)
+    encryptor     = ActiveSupport::MessageEncryptor.new(secret, cipher: @encrypted_cookie_cipher, serializer: @serializer)
+
+    begin
+      encryptor.decrypt_and_verify(@cookie)
+    rescue Exception => error
```
piotrekpasciak: We should  at all cost avoid using rescue on `Exception` or `StandardError` classes. Can you catch more specific error for `.decrypt_and_verify(@cookie)` method?

app/controllers/api/session_helper_controller.rb
```diff
@@ -0,0 +1,17 @@
+require 'cgi'
+require 'active_support'
+
+class Api::SessionHelperController < Api::BaseController
+  respond_to :json
+
+  def check_login
```
piotrekpasciak: Why do we have to follow this logic instead of checking `current_participant` presence? What's the business logic behind?

config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_aicrowd_session'
+Rails.application.config.session_store :cookie_store, key: '_aicrowd_session', domain: :all
```
piotrekpasciak: What's the purpose of `domain: :all`? Where do you need to handle subdomains?

piotrekpasciak:
Please write also basic test coverage for:
- endpoint
- ServiceObject class

Example of endpoint spec:
https://github.com/AIcrowd/AIcrowd/pull/1289/files#diff-0c3018a23463f3bd5e7ab821de45ac7dR16

app/controllers/api/session_helper_controller.rb
```diff
@@ -0,0 +1,17 @@
+require 'cgi'
```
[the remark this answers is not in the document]
alkeshrails: Done!

piotrekpasciak:
When I was testing `current_participant` I was getting CORS errors on staging.

app/services/session_service.rb
alkeshrails: Didn't specific issue.

app/controllers/api/session_helper_controller.rb
alkeshrails: Thanks

app/controllers/api/session_helper_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'active_support'
+
+class Api::SessionHelperController < Api::BaseController
```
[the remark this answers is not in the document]
alkeshrails: Actually all current API is in api folder so I create file according to them.

app/controllers/api/session_helper_controller.rb
alkeshrails: Will check together.

app/controllers/api/session_helper_controller.rb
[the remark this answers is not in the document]
piotrekpasciak: Not anymore. Take a look here:
https://github.com/AIcrowd/AIcrowd/tree/master/app/controllers/api/v1/challenges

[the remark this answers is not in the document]
piotrekpasciak: You're still inheriting from wrong controller.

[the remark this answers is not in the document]
piotrekpasciak: Please write test coverage at least for API endpoint. Request specs are recommended.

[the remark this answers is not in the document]
alkeshrails: Here i move api into v1.
for test cases I need to manage domain of aicrowd in spec.

PR #1299, opened 2020-03-31
Handle invalid credentials in Discourse services

piotrekpasciak:


app/services/discourse/create_category_service.rb
```diff
@@ -26,6 +26,8 @@ def call
       Logger.new(::Discourse::BaseService::LOGGER_URL).error("##{challenge.id} - Unable to create category - #{e.message}")
 
       failure('Discourse API is unavailable')
+    rescue Discourse::UnauthenticatedError, Discourse::NotFoundError => e
```
skbly7, line 29: Can't we keep it at one place v/s duplicating rescue everywhere?

Example:

`create_category_request` returns back with two params, `success` and `data`. In case `success` is `false` data contains the error message, and when `true` contains the data we are getting right now.

Similarly for all the DIscourse APIs.

If tomorrow, new Discourse API is implemented, lot of rescue, etc are needed every time which isn't a good thing to have.

piotrekpasciak: I don't understand the example and meaning of it, but I'll make code more DRY.

piotrekpasciak: @skbly7 I've updated the PR. Please take a look.

PR #1304, opened 2020-04-02
Added reCaptcha during signups to prevent bot signups

skbly7:


app/controllers/participants/registrations_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Participants::RegistrationsController < Devise::RegistrationsController
+  include Recaptcha
+  prepend_before_action :check_captcha, only: [:create] # Change this to be any actions you want to protect.
+
+  private
+    def check_captcha
+      unless !Rails.env.production? || verify_recaptcha
```
piotrekpasciak, line 9: In Ruby we have concept of guard clauses. They make code mutch easier to read:
https://github.com/rubocop-hq/ruby-style-guide#no-nested-conditionals

You can change first line to `unless !Rails.env.production? || verify_recaptcha` and make all other code out of this block.

skbly7: Thanks, updated the code based on the suggestion.

PR #1308, opened 2020-04-02
Allow organizers to change challenge status + refactor create submission button policy

nemaniarjun:


app/policies/challenge_policy.rb
```diff
@@ -103,19 +103,24 @@ def has_accepted_participation_terms?
...
-      if @record.clef_challenge.present?
-        if clef_participant_registered?(@record)
-          return true # return true if running and clef challenge and registered
-        else
-          return false # return false if running and clef_challenge and NOT REGISTERED
-        end
-      else
-        return true # return true if running and no clef challenge
-      end
+    if @record.completed? # if the challenge is over
```
piotrekpasciak, line 108: You can these 2 lines to:
return false if @record.completed? && !@record.post_challenge_submissions?

app/policies/challenge_policy.rb
```diff
@@ -103,19 +103,24 @@ def has_accepted_participation_terms?
...
+      return false unless @record.post_challenge_submissions?
     end
-    return false # no positive condition met
+
+    # If we reach this part that means that the challenge status is set to 'running'
+    # return true if not a clef challenge
+    return true unless @record.clef_challenge.present?
```
skbly7, line 114: This is wrong, it shouldn't return `true` here but check for `clef_participant_registered` as well.

app/policies/challenge_policy.rb
nemaniarjun: noted and making the change

app/policies/challenge_policy.rb
skbly7: Ignore, my bad.

app/policies/challenge_policy.rb
```diff
@@ -103,19 +103,22 @@ def has_accepted_participation_terms?
   def submissions_allowed?
     return false unless @record.online_submissions
     return true if edit?
+    return false if @record.starting_soon? || @record.draft?
```
akshatcx, line 106: We might want to make test submissions(internal) to ensure the evaluator is working. (when the challenge is in draft mode)

nemaniarjun: The line above this will show the button to anyone who has edit access i.e. admins and organizers.

akshatcx: oh okay :+1:

nemaniarjun:
This was pushed directly to the relevant branch.

PR #1310, opened 2020-04-02
Remove social link to leaderboard index

alkeshrails:


skbly7:
LGTM, we are disabling this new feature for now and lets polish it before making public! +1

PR #1311, opened 2020-04-02
Fix og tags for videos

alkeshrails:


app/helpers/application_helper.rb
```diff
@@ -99,7 +99,7 @@ def themed_button(opts = {})
 
   def social_share(site, challenge, mediable)
     if mediable.class.name == 'Leaderboard'
-      data_url = "#{request.url}"
+      data_url = "#{request.base_url}/challenges/#{challenge.friendly_id}/submissions/#{mediable.submission.id}"
```
piotrekpasciak, line 102: We should use Rails helpers whenever possible. I think you meant to use something like `challenge_submission_url(challenge, submission)`

app/views/layouts/application.html.erb
```diff
@@ -10,7 +10,7 @@
     <meta property="og:title" content="<%= meta_title %>" />
     <meta property="og:type" content="website" />
     <meta property="og:url" content="<%= request.original_url %>" />
-    <meta property="og:image" content="<%= meta_image %>" />
+    <meta property="og:video:secure_url" content="<%= meta_video %>" />
```
skbly7, line 13: Why are we getting rid of `og:image`? It doesn't seem right.

PR #1314, opened 2020-04-03
Set header for challenge tab

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/517801550

### DIFF
--- a/app/helpers/meta_tags_helper.rb
+++ b/app/helpers/meta_tags_helper.rb
@@ -4,13 +4,13 @@ module MetaTagsHelper
       "AIcrowd | #{@challenge.challenge} | #{controller_name.capitalize}"
     elsif show_action?
       if controller_name == 'challenges'
-        "AIcrowd | " + @challenge.challenge
+        "AIcrowd | " + @challenge.challenge + "| #{controller_name.capitalize}"
       elsif controller_name == 'participants'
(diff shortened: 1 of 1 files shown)


app/helpers/meta_tags_helper.rb
```diff
@@ -1,14 +1,16 @@
...
+        "AIcrowd | " + @challenge.challenge
       elsif controller_name == 'participants'
-        @participant.slug
+        "AIcrowd | " + @participant.slug
       elsif controller_name == 'organizers'
-        @organizer.slug
+        "AIcrowd | " + @organizer.slug
       elsif controller_name == 'submission'
-        @challenge.challenge
+        "AIcrowd | " + @challenge.challenge
```
skbly7, line 13: Can be `AIcrowd | " + @challenge.challenge + " | Submissions"` and so on. Basically someone should be able to know which tab/page it is using just the page title.

alkeshrails: Actually here, in some cases `show_action?` will false like leaderboard tab call index action, so I manage all cases of challenges in first condition https://github.com/AIcrowd/AIcrowd/pull/1314/files#diff-c44a693369f9c1850f1441d029f489daR3
As requested I also add `controller_name`

PR #1315, opened 2020-04-06
Fix leaderboard social share

alkeshrails:


app/helpers/application_helper.rb
```diff
@@ -97,31 +97,31 @@ def themed_button(opts = {})
...
 
-  def social_share(site, challenge, mediable)
-    if mediable.class.name == 'Leaderboard'
-      data_url = "#{request.url}"
-      img = having_media_thumbnail_image?(challenge, mediable) ? s3_public_url(mediable, :thumb) : nil
-    elsif mediable.class.name == 'Submission'
-      data_url = "#{request.base_url}/#{mediable.short_url}" if mediable.class.name == 'Submission'
-      img = having_media_large_image?(challenge, mediable) ? s3_public_url(mediable, :large) : nil
+  def social_share(site, challenge, mediable, text=nil)
+    title = challenge.challenge
```
piotrekpasciak, line 101: Would be cleaner to:
`title = text.presence || challenge.challenge`

app/helpers/application_helper.rb
```diff
@@ -149,4 +149,28 @@ def banner_color
   def footer_text
     raw Setting.footer_record
   end
+
+  def challenge_submission_url(challenge, submission)
+    "#{request.base_url}/challenges/#{challenge.friendly_id}/submissions/#{submission.id}"
```
piotrekpasciak, line 154: We should use Rails helpers to generate links as it's easier to maintain code with them:
`challenge_submission_url(challenge_id:  challenge.friendly_id, id: submission.id)`

app/helpers/meta_tags_helper.rb
```diff
@@ -57,6 +57,16 @@ def meta_image
     end
   end
 
+  def meta_video
+    if show_action? && controller_name == 'submissions'
```
piotrekpasciak, line 61: It's bad practice to use instance variables from view in helper. Please pass `challenge` and `submission` as params of this method.

skbly7: This is being used in `application/layout.yml` and we may or may not have variables in scope I guess. (I followed the other functions in `meta_tags_helper.rb` due to this reason)

PR #1317, opened 2020-04-06
Adding blog link in the leaderboard page

nikhilrayaprolu:
image

### DIFF
--- a/app/views/rating_leaderboard/index.html.erb
+++ b/app/views/rating_leaderboard/index.html.erb
@@ -3,6 +3,7 @@
     <div class="row">
       <div class="col-12">
         <h1>AIcrowd Rankings</h1>
+        <%= link_to "Learn more about rankings", blog_path(10) %>
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


app/views/rating_leaderboard/index.html.erb
```diff
@@ -3,6 +3,7 @@
     <div class="row">
       <div class="col-12">
         <h1>AIcrowd Rankings</h1>
+        <%= link_to "Learn more about rankings", blog_path(10) %>
```
skbly7, line 6: Refactor it to link blog via slug, so we have control in future. Linking via ID isn't a good way to do so.

skbly7: Ideally, have that slug with some info about rating added in seed as well. (extended goal)

piotrekpasciak: Like @skbly7 mentioned. Using record `id` is very bad practice.

nikhilrayaprolu:
@skbly7 @piotrekpasciak made requested changes

PR #1328, opened 2020-04-09
fixing the decay issues

nikhilrayaprolu:


app/services/user_rating_service.rb
```diff
@@ -1,5 +1,6 @@
 class UserRatingService
   def initialize(round_id)
+    attr_accessor :temporary
```
piotrekpasciak, line 3: attr_accessor should be in class, not in inside method.

PR #1329, opened 2020-04-09
Add Api::V1::Challenges#create and Api::V1::Challenges#update  endpoints

piotrekpasciak:
Questions:
- Should this endpoint accept json_file or json_string?
- Where is this endpoint going to be hit from? From web app front-end or foreign server? ( I have to choose what kind of authentication I have to use for it)

Shivam answers:
- We will use json_string
- It's going to be hit from forieng_server, use API-key authentication

Screenshot 2020-04-13 at 22 50 14


app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  module V1
+    module Challenges
+      class ChallengesController < ::Api::BaseController
+        before_action :auth_by_admin_api_key, only: [:create, :update]
```
skbly7, line 5: Hey, we need to start [using participant's key as auth](https://github.com/AIcrowd/AIcrowd/blob/master/app/models/participant.rb#L3) here instead of admin key. And that key can be used as follows: 

```
key -> participant -> can_edit? -> allow creation or update to challenge.
```

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        before_action :set_challenge, only: :update
+
+        def create
+          result = ::Challenges::ImportService.new(import_params: params, organizers: []).call
```
skbly7, line 9: Let's make it mandatory to provide at least one organizer when creating the challenge.  Otherwise, permission management etc can be tricky later on.

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+          if result.success?
+            head :created
```
skbly7, line 12: Let's return the challenge slug, id, title, URL to challenge (few params), so the clients can know about new challenge programmatically.

piotrekpasciak:
@skbly7 I've applied your remarks to this PR.

app/controllers/api/v1/base_controller.rb
```diff
@@ -1,22 +1,34 @@
...
       include Pundit
 
       rescue_from Pundit::NotAuthorizedError, with: :not_authorized_message
 
       protected
 
+      def auth_by_participant_api_key
+        authenticate_or_request_with_http_token do |token, options|
+          @api_user = Participant.find_by(api_key: token)
+          (token == ENV['CROWDAI_API_KEY'] || @api_user.present?)
```
skbly7, line 14: Hey, this function is perfect now, a minor suggestion is to use @api_user as site-admin in case `token == ENV['CROWDAI_API_KEY']` (say `Users.find(admin=true)[0]`). It will be easier to handle authentication wherever you are using @api_user variable later on.

app/serializers/api/v1/challenge_serializer.rb
```diff
@@ -0,0 +1,22 @@
+module Api
```
skbly7, line 1: 💯

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  module V1
+    module Challenges
+      class ChallengesController < ::Api::V1::BaseController
+        before_action :auth_by_participant_api_key, only: [:create, :update]
+        before_action :set_challenge, only: :update
+
+        def create
+          result = ::Challenges::ImportService.new(
+            import_params: params,
+            organizers:    api_user&.organizers.to_a
```
skbly7, line 11: Looks like we are settings challenge organizers as all the organizers the authenticated user is part of? 

If yes, it needs to be something like...
Get the list of organizers provided in JSON data, check the following conditions:
1. At least one organizer is provided in JSON
2. The authenticated user is part/access to of all the organizers provided

piotrekpasciak: I was reusing old export / import functionality. It was setting organizers based on session.

I will update it to support setting organizers based on json string.

app/models/participant.rb
```diff
@@ -107,6 +107,10 @@ class Participant < ApplicationRecord
             length:      { in: 2...100 },
             allow_blank: true
 
+  def self.api_admin
```
skbly7, line 110: This seems to be part of commit by mistake?

piotrekpasciak: Nope, it's a class method that holds admin_user record to set as `api_user` whenever ENV[
'CROWDAI_API_KEY'] was used as token.

skbly7: Ok, then instead of hard coding email ID (and hoping this record exist), let's use something like `find(admin:true)` first record maybe?

skbly7: Discussed offline, we are going ahead with taking API admin username as ENV variable, so it points to single user but at the same time isn't hardcoded.

piotrekpasciak: I've added discussed ENV variables.

PR #1330, opened 2020-04-09
Update Challenges::ImportService to support updating of existing challenges

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/523048372

Screenshot 2020-04-14 at 13 09 36


app/models/concerns/challenges/import_constants.rb
```diff
@@ -98,26 +98,6 @@ module ImportConstants
         :additional_checkbox_text,
         :additional_checkbox_text_markdown,
         :version
-      ],
```
skbly7, line 101: Q: Hi, does this change mean people can't add/remove ChallengeRound via API? We will need this functionality to keep on working either via upsert or different API call.

piotrekpasciak: Unfortunately following line doesn't allow to recreate challenge_rounds:
https://github.com/AIcrowd/AIcrowd/blob/master/app/models/challenge_round.rb#L5

Will implement separate endpoints for challenge_rounds

PR #1333, opened 2020-04-10
set domain for config session

alkeshrails:


config/initializers/session_store.rb
```diff
@@ -1,3 +1,7 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_aicrowd_session'
+if Rails.env.development? || Rails.env.test?
+  Rails.application.config.session_store :cookie_store, key: '_aicrowd_session'
+else
+  Rails.application.config.session_store :cookie_store, key: '_aicrowd_session', domain: ENV['DOMAIN_NAME'].gsub("https://www.", "").gsub("https://", "")
```
skbly7, line 6: Please take care of `http://` too here as we discussed earlier!

alkeshrails: @skbly7 made suggested changes, please have a look.

PR #1340, opened 2020-04-14
Add service to get baseline discourse and show on challenge overview

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/522785861

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -168,7 +168,7 @@ class Challenge < ApplicationRecord
 
   def baseline_discussion
     discource_baselines = Discourse::FetchBaselineTagService.new(challenge: self).call
-    discource_baselines.value.first(5) if discource_baselines[:success]
+    discource_baselines.value.first(5) if discource_baselines.success?
   end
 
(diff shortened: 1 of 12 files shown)


app/services/discourse/fetch_baseline_tag_service.rb
```diff
@@ -0,0 +1,28 @@
+module Discourse
+  class FetchBaselineTagService < ::Discourse::BaseService
```
piotrekpasciak, line 2: Could you write basic test coverage for this class? You'll probably need to use VCR.

app/models/challenge.rb
```diff
@@ -163,6 +163,11 @@ def current_challenge_rules
     ChallengeRules.where(challenge_id: id).order('version DESC').first
   end
 
+  def baseline_discussion
+    discource_baselines = Discourse::FetchBaselineTagService.new(challenge: self).call
+    discource_baselines.value.first(5) if discource_baselines[:success]
```
piotrekpasciak, line 168: You can use `discource_baselines.success?` here instead.

PR #1344, opened 2020-04-16
moving the submission badges into a helper

nikhilrayaprolu:


app/helpers/badges_helper.rb
```diff
@@ -0,0 +1,20 @@
+module BadgesHelper
+  def submission_badges(participant_id)
```
piotrekpasciak, line 2: I think this should be actually under model/concerns directory. Helpers in Rails are associated with methods that are used in HTML views and that's why it's not a good idea to put method like this inside of helper.

piotrekpasciak, line 2: `create_submission_badges` would be a better name for this method.

piotrekpasciak: It would be a good idea to write test coverage for this method.

app/models/concerns/submission_badges.rb
```diff
@@ -0,0 +1,20 @@
+module SubmissionBadges
+  def create_submission_badges(participant_id)
+    if grading_status_cd == 'graded'
+      participant_submission_count = Submission.where(participant_id: participant_id, grading_status_cd: 'graded').count
+      if participant_submission_count == 10
+        Participant.find_by(id: participant_id).add_badge(9)
```
skbly7, line 6: Add badges by using their name instead of id, otherwise code isn't readable in long term.

piotrekpasciak: Additionally using ids is very unreliable and may generate bugs that are very difficult to track and fix.

PR #1345, opened 2020-04-16
Ckeditor integration with ActiveAdmin markdown fields

alkeshrails:
EditJobPosting
 Integrate Ckeditor for ActiveAdmin
blogs, column body
challenge_calls, column description
job_postings, column description
participation_terms, column terms


db/migrate/20200415125555_remove_column_markdown.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveColumnMarkdown < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :blogs, :body_markdown, :text
```
piotrekpasciak, line 3: Are you sure you need to only remove these fields? Don't you need to migrate value from `body_markdown`  to `body` field etc?

You can use rake task to migrate the data.

skbly7: It is safe to delete `body_markdown` because HTML content for the same is present in `body` field.

PR #1348, opened 2020-04-16
Add Discourse::CreateGroupService

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/522768303

### DIFF
--- a/app/services/discourse/create_group_service.rb
+++ b/app/services/discourse/create_group_service.rb
@@ -1,0 +1,42 @@
+module Discourse
+  class CreateGroupService < ::Discourse::BaseService
+    def initialize(challenge:)
+      @client    = prepare_http_client
+      @challenge = challenge
+    end
+
+    def call
+      retry_count ||= 0
+
+      return failure('Discourse API client couldn\'t be properly initialized.') if client.nil?
+
+      with_discourse_errors_handling do
+        response = create_group_request(ensure_uniqueness: retry_count.positive?)
+
+        challenge.update!(discourse_group_id: response.body['basic_group']['id'])
+
+        success(response)
+      rescue Discourse::UnprocessableEntity => e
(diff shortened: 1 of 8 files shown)


app/services/discourse/create_group_service.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        challenge.update!(discourse_group_id: response.body['basic_group']['id'])
+
+        success(response)
+      rescue Discourse::UnprocessableEntity => e
+        raise e if retry_count.positive?
+
+        retry_count += 1
+
+        retry if e.message.include?('Name has already been taken')
```
skbly7, line 24: We are not re-generating slug, thats why in case it fail due to this error, it would keep on retrying forever?

app/services/discourse/create_group_service.rb
```diff
@@ -0,0 +1,42 @@
...
+  class CreateGroupService < ::Discourse::BaseService
+    def initialize(challenge:)
+      @client    = prepare_http_client
+      @challenge = challenge
+    end
+
+    def call
+      retry_count ||= 0
+
+      return failure('Discourse API client couldn\'t be properly initialized.') if client.nil?
```
skbly7, line 11: Why we are needed to have this condition in every new service file? Can this be moved to BaseService and be there only somehow?

app/services/discourse/create_group_service.rb
piotrekpasciak: 2 lines above `raise e if retry_count.positive?` that should prevent infinite loop of retrying.

app/services/discourse/create_group_service.rb
piotrekpasciak: I've handled duplication in this PR:
https://github.com/AIcrowd/AIcrowd/pull/1363

PR #1352, opened 2020-04-17
Add Discourse::AddUsersToGroupService

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/523058269

### DIFF
--- a/app/services/discourse/add_users_to_group_service.rb
+++ b/app/services/discourse/add_users_to_group_service.rb
@@ -1,0 +1,40 @@
+module Discourse
+  class AddUsersToGroupService < ::Discourse::BaseService
+    def initialize(challenge:, participants:)
+      @client       = prepare_http_client
+      @challenge    = challenge
+      @participants = participants
+    end
+
+    def call
+      with_discourse_errors_handling do
+        return failure('Discourse group doesn\'t exist in our database') if challenge.discourse_group_id.blank?
+        return failure('You need to provide at least one participant') if participants.blank?
+
+        response = add_users_to_group_request
+
+        success(response)
+      rescue Discourse::UnprocessableEntity => e
+        return failure(e.message) if e.message.include?('is already a member of this group')
+
+        raise e
+      end
+    end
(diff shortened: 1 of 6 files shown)


app/services/discourse/add_users_to_group_service.rb
```diff
@@ -0,0 +1,41 @@
+module Discourse
+  class AddUsersToGroupService < ::Discourse::BaseService
```
skbly7, line 2: Where is this service being called from?

app/services/discourse/create_group_service.rb
```diff
@@ -0,0 +1,42 @@
+module Discourse
+  class CreateGroupService < ::Discourse::BaseService
```
skbly7, line 2: This seems to be already going as part of #1348 PR?

app/services/discourse/add_users_to_group_service.rb
```diff
@@ -0,0 +1,41 @@
...
+    def add_users_to_group_request
+      client.put(
+        "/groups/#{challenge.discourse_group_id}/members.json", {
+          usernames: participants_usernames
+        }
+      )
+    end
+
+    def participants_usernames
+      participants.pluck(:name).join(',')
```
skbly7, line 38: `@participants` or `participants` ?

app/services/discourse/add_users_to_group_service.rb
piotrekpasciak: I don't call it yet. I'll implement business logic in later PR's.

app/services/discourse/create_group_service.rb
piotrekpasciak: I've rebased with master and resolved the duplication

app/services/discourse/add_users_to_group_service.rb
piotrekpasciak: `participants` - I have attr_reader defined above.

PR #1356, opened 2020-04-18
Feature/improve filter url

alkeshrails:
Readable challenge filter.
Screenshot from 2020-04-18 22-49-52

If there is a space between category tag, than it will show %20.

latest review app :
Screenshot from 2020-04-23 15-32-05

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -76,6 +76,10 @@ class Challenge < ApplicationRecord
               ELSE 5
             END, challenges.participant_count DESC")
   end
+  scope :prize_cash, -> { where.not(prize_cash: nil) }
+  scope :prize_travel, -> { where.not(prize_travel: nil) }
+  scope :prize_academic, -> { where.not(prize_academic: nil) }
+  scope :prize_misc, -> { where.not(prize_misc: nil) }
 
   after_initialize :set_defaults
   after_create :create_discourse_category
@@ -215,21 +219,4 @@ class Challenge < ApplicationRecord
     return unless active_round
     active_round.leaderboards.limit(5)
   end
-
-  def self.prize_cash
-    where.not(prize_cash: nil)
-  end
-
-  def self.prize_travel
-    where.not(prize_travel: nil)
-  end
-
-  def self.prize_academic
-    where.not(prize_academic: nil)
-  end
-
-  def self.prize_misc
-    where.not(prize_misc: nil)
-  end
 end
-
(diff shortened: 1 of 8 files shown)


app/models/challenge.rb
```diff
@@ -197,4 +197,20 @@ def other_scores_fieldnames_array
     arr = other_scores_fieldnames
     arr&.split(',')&.map(&:strip) || []
   end
+
+  def self.prize_cash
```
piotrekpasciak, line 201: These should be scopes and probably named like `not_prize_cash`.

app/services/challenges/filter_service.rb
```diff
@@ -8,15 +8,15 @@ def initialize(params, challenges)
...
+      @challenges = @challenges.where(status_cd: @params['status']) if @params['status'].present?
       # prize filter
-      @challenges = @challenges.where.not(prize_cash: nil) if @params.dig(:prize, 'prize_type')&.include?('prize_cash')
-      @challenges = @challenges.where.not(prize_travel: nil) if @params.dig(:prize, 'prize_type')&.include?('prize_travel')
-      @challenges = @challenges.where.not(prize_academic: nil) if @params.dig(:prize, 'prize_type')&.include?('prize_academic')
-      @challenges = @challenges.where.not(prize_misc: nil) if @params.dig(:prize, 'prize_type')&.include?('prize_misc')
-      # return filtered challenges
+      if @params['prizes'].present?
+        @params['prizes'].split(',').each do |prize|
+          @challenges = @challenges.send(prize)
```
piotrekpasciak, line 17: Please use `public_send` when doing meta-programming. Send can call private methods what should never be allowed.

app/views/shared/challenges/_filter.html.erb
```diff
@@ -7,42 +7,75 @@
...
           <% end %>
         </div>
       <% end %>
     </div>
   </div>
-  <%= submit_tag 'Apply', class: 'btn btn-link btn-sm apply-filter-btn' %>
+  <%= f.submit 'Apply', name: nil, class: 'btn btn-link btn-sm apply-filter-btn' %>
 <% end %>
+<script type="text/javascript">
+  $("form").submit(function(){
```
piotrekpasciak, line 50: This script should be moved to StimulusJS controller.

app/views/shared/challenges/_filter.html.erb
```diff
@@ -7,42 +7,75 @@
...
+    event.preventDefault();
+
+    categories = [];
```
piotrekpasciak, line 53: Shouldn't you declare these variables with `let`?

PR #1357, opened 2020-04-19
removing the admin ratings

nikhilrayaprolu:


app/views/participants/_show_temporary_reputation.erb
```diff
@@ -1,9 +1,5 @@
 <span class="reputation-info" style="<%= style ||= '' %>">
-<% if participant.admin %>
-  <span class="reputation-score" data-toggle="tooltip" title="Admin">
-    ###
-  </span>
-<% elsif current_participant&.admin? %>
+<% if current_participant&.admin? && !participant.admin %>
```
piotrekpasciak, line 2: What's the purpose of `&& !participant.admin` ?

Previously this check was not needed. Are you sure you need it now?

nikhilrayaprolu: previously we had if else condition, but we only need the else condition and need to show nothing in `if` condition. So just added !participant.admin (if condition part into the else and made else to if)

PR #1359, opened 2020-04-19
badges and rating fixes

nikhilrayaprolu:


app/jobs/discourse_badges_daily_job.rb
```diff
@@ -1,16 +1,20 @@
 class DiscourseBadgesDailyJob < ApplicationJob
   queue_as :default
+  def get_badges()
```
piotrekpasciak, line 3: If it's not used outside of this class, please move it to `private` section.

piotrekpasciak, line 3: In Ruby there is convention to not use `()` when function doesn't take arguments. Please remove `()`.

app/jobs/discourse_badges_daily_job.rb
```diff
@@ -1,16 +1,20 @@
...
+  end
   def perform(*args)
     discourse_initial_id = ENV['DISCOURSE_INITIAL_ID'].to_i
-    previous_id = DiscourseUserBadgesMetum.order(:created_at).last&.previous_id.to_i
-    user_badges = Discourse::FetchBadgesService.new(previous_id).call.value
-    user_badges.each do |user_badge|
-      participant = Participant.find_by(name: user_badge['username'])
-      if participant
-        participant.add_badge(discourse_initial_id.to_i + user_badge['badge_id'].to_i)
+    while user_badges = get_badges && user_badges.present? do
```
piotrekpasciak, line 9: Please assign `user_badges` outside of loop. It will be mutch easier to read it.

piotrekpasciak: This `while` loop is quite unusual. Are you sure we need. it? Can't we simply write   some kind of `if` statement like `return if user_badges.blank?` ?

app/jobs/rollback_badges_job_job.rb
```diff
@@ -0,0 +1,7 @@
+class RollbackBadgesJobJob < ApplicationJob
```
piotrekpasciak, line 1: Please don't add empty classes to repository.

config/initializers/devise.rb
```diff
@@ -125,7 +125,7 @@
   # initial account confirmation) to be applied. Requires additional unconfirmed_email
   # db field (see migrations). Until confirmed, new email is stored in
   # unconfirmed_email column, and copied to email column on successful confirmation.
-  config.reconfirmable = true
+  config.reconfirmable = false
```
piotrekpasciak, line 128: Why did you change it here? This has serious consequences for participants that change e-mails.

app/jobs/discourse_badges_daily_job.rb
nikhilrayaprolu: Yes, since the discourse API is not sending records more than 840, I have to loop over the API, until the all rows are exhausted.

Also if I don't put the condition in the while loop, it doesn't follow DRY, I have to repeat the same line before the loop is called and at the end of loop.

nikhilrayaprolu: Any suggestions for the above two issues?

config/initializers/devise.rb
nikhilrayaprolu: image
Looks like this is what causing the Mandrill Error,

There I am updating the Participants model
by calling `Participant.update`
It's leading to the Mandrill call back being called, So the only fix I found is to do this.

One other fix would be to do `participant.skip_confirmation_notification!` but it doesn't work here in this case because, I am doing a bulk update with multiple participant keys passed to update call.

nikhilrayaprolu: Any other suggestion if you don't want me to do it this way?

skbly7: @nikhilrayaprolu You can remove this change from your commit. I suggested this change to exist so you can test properly in review app (and then remove it later on).

A proper fix for the same is going via @piotrekpasciak's PR here: https://github.com/AIcrowd/AIcrowd/pull/1362

app/jobs/discourse_badges_daily_job.rb
piotrekpasciak: What about following approach?
`user_badges = get_badges
while user_badged.present? do
  business_logic...

  user_badges = get_badges
end`

piotrekpasciak: I think that would be DRY enough.

nikhilrayaprolu:
@piotrekpasciak @skbly7 made the requested changes.

PR #1360, opened 2020-04-19
Bugfix: Invalid Submission on challenge rules acceptance

skbly7:
`accepted_dataset_toc` refactoring wasn't complete due to which "Invalid Submission" occurs on every new submission.

This commit changes the condition to challenge rules policy. Along with it moved from `initialize` to `call` function, so return false works properly and doesn't continue working.

Condition for `ready` exist, so this condition is checked only on new submission and not recalculations.


app/services/grader_service.rb
```diff
@@ -76,6 +63,16 @@ def api_query
     participant    = @submission.participant
     submission_key = get_submission_key
     team_id        = participant.teams.where(challenge: challenge).first&.id || 'undefined'
+    challenge_participant = challenge.challenge_participants.find_by(participant_id: participant.id)
+
+    # The participation terms condition should only be checked on submission creation not recomputes
+    if @submission.grading_status == 'ready' && (challenge_participant.blank? || !challenge.has_accepted_challenge_rules?(participant))
+      Submission.update(
```
piotrekpasciak, line 70: `update!` instead of `update`

PR #1361, opened 2020-04-20
Submission Flow v2 (+ ACE,  + EvaluationsAPI)

skbly7:
This is a comparatively bigger PR and contains the following changes:

1. New service & job modification to get submissions evaluated by Evaluations API
2. Completion of ACE editor integration
3. Submission flow modification, starting to provide multiple options for submissions via the web interface.


app/views/submissions/new/_code.html.erb
```diff
@@ -0,0 +1,95 @@
...
+          </select>
+        </div>
+      </div>
+      <textarea id='ace-editor'></textarea>
+      <%= f.hidden_field :submission_data, id: 'ace-editor-value', value: @form %>
+      <div class="ace-editor-pad"></div>
+    </div>
+    <small class="form-text">The code should generate all the artifacts, see example codes for more idea.</small>
+
+    <script>
```
skbly7, line 48: TODO: Migrate JS code from views

app/views/submissions/new/_gitlab.html.erb
```diff
@@ -0,0 +1,113 @@
...
+    <p>You now should be able to see the details of your submission in the repository's issue tracker. Example based on above: <a href="https://gitlab.aicrowd.com/<%= current_participant.name %>/<%= @challenge.slug %>/issues" target="_blank">https://gitlab.aicrowd.com/<%= current_participant.name %>/<%= @challenge.slug %>/issues</a></p>
+
+    <p>Happy Submitting!! 🚀</p>
+    </div>
+  </div>
+  <div class="col-lg-5">
+    <div class="md-content">
+      <blockquote>💬 Any questions? Discuss with community and work together by joining <a href="i<%= challenge_discussion_path(challenge) =>">Discourse Forum</a>.</blockquote>
+    </div>
+    <img class="img-fluid mb-4" src="https://i.imgur.com/o7Zkjl3.png" />
```
skbly7, line 105: TODO: Upload imgur images to codebase after approval

config/initializers/evaluations_api.rb
```diff
@@ -0,0 +1,8 @@
+require 'aicrowd_evaluations'
+
+AIcrowdEvaluations.configure do |config|
+  config.api_key['AUTHORIZATION'] = ENV["EVALUATIONS_API_KEY"]
+  config.host = ENV["EVALUATIONS_API_HOST"]
+  config.verify_ssl = false
```
skbly7, line 6: TODO: Make verify_ssl=true after production API's SSL cert issue is solved

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+    create_submission(grader.id)
+  rescue StandardError => e
+    Submission.update(
+      @submission.id,
+      grading_status:  'failed',
+      grading_message: e.message)
+    raise e
+  end
+
+  def create_submission(grader_id)
```
piotrekpasciak, line 20: Please move everything that's not used outside of class to private section.

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+  def initialize(submission_id:)
+    @submission = Submission.find(submission_id)
+    @grader_id = @submission.challenge.challenge_client_name.to_i
+  end
+
+  def call
+    grader = get_grader
+    create_submission(grader.id)
+  rescue StandardError => e
+    Submission.update(
```
piotrekpasciak, line 13: We need to follow policy of failing early to make system predictable. Use `update!` here instead.

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+class EvaluationsApiService
+  def initialize(submission_id:)
+    @submission = Submission.find(submission_id)
+    @grader_id = @submission.challenge.challenge_client_name.to_i
+  end
+
+  def call
+    grader = get_grader
+    create_submission(grader.id)
+  rescue StandardError => e
```
piotrekpasciak, line 12: We need to avoid using `rescue StandardError => e` at all cost. It catches all possible errors and makes system unpredictable.

Please try to list errors you actually want to catch.

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+    payload = AIcrowdEvaluations::Submissions.new attribute
+    submission_api = AIcrowdEvaluations::SubmissionsApi.new
+    submission_response = submission_api.create_submission(payload)
+    Submission.update(
+      @submission.id,
+      grading_status: 'submitted',
+      grading_message: 'Evaluating...')
+  rescue StandardError => e
```
piotrekpasciak, line 39: Like in comment above.

Btw. there is no need to catch error twice. You can catch them all in `call` method.

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      @submission.id,
+      grading_status:  'failed',
+      grading_message: e.message)
+    raise e
+  end
+
+  def create_submission(grader_id)
+    artifact = get_submission_artifact
+    attribute = {
+      :participant_id => @submission.participant.id,
```
piotrekpasciak, line 23: Use new hash syntax.

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      :submission_data => {
+        :type => artifact.submission_type,
+        # TODO: Directly submit s3 url as submission instead of downloading
+        :code => download_s3_file(artifact.submission_file_s3_key)
+      }
+    }
+    payload = AIcrowdEvaluations::Submissions.new attribute
+    submission_api = AIcrowdEvaluations::SubmissionsApi.new
+    submission_response = submission_api.create_submission(payload)
+    Submission.update(
```
piotrekpasciak, line 35: `update!`

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      @submission.id,
+      grading_status: 'submitted',
+      grading_message: 'Evaluating...')
+  rescue StandardError => e
+    Submission.update(
```
piotrekpasciak, line 40: `update!`

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
...
+    raise e
+  end
+
+  def get_grader
+    # TODO: Use the client names instead of grader_id
+    grader_api = AIcrowdEvaluations::GradersApi.new
+    grader = grader_api.get_grader(@grader_id)
+    if grader.status != "Completed"
+      message = "Grader not ready for submissions"
+      Submission.update(
```
piotrekpasciak, line 53: `update!`

app/services/evaluations_api_service.rb
```diff
@@ -0,0 +1,69 @@
+require 'aicrowd_evaluations'
+
+class EvaluationsApiService
```
piotrekpasciak, line 3: It would good idea to put it under namespace e.g `Evaluations` and change name of the class.

Something like this would tell mutch more about purpose of this class:

module Evaluations
  class CreateSubmissions
  end
end

piotrekpasciak: We could use test coverage for large class like that.

app/views/submissions/new/_code.html.erb
piotrekpasciak: I think it wouldn't be mutch work to move it to StimulusJS controller.

config/application.yml.example
```diff
@@ -1,6 +1,8 @@
 AWS_REGION: "REDACTED"
 AICROWD_API_KEY: '8c69ef80a5eb4438ec326d200ce6ab39'
 AICROWD_API_EMAIL: 'api_admin@example.com'
+EVALUATIONS_API_KEY: "REDACTED"
```
piotrekpasciak, line 4: I think you should add these to VCR config like that:
https://github.com/AIcrowd/AIcrowd/blob/master/spec/support/config/vcr.rb#L13

db/migrate/20200330140331_add_code_submissions_to_challenge.rb
```diff
@@ -0,0 +1,5 @@
+class AddCodeSubmissionsToChallenge < ActiveRecord::Migration[5.2]
```
piotrekpasciak, line 1: I think it would be a good idea to remove these migrations from this PR:
db/migrate/20200420071527_remove_code_submissions_from_challenges.rb
db/migrate/20200330140331_add_code_submissions_to_challenge.rb
db/migrate/20200420065623_remove_code_data_from_submissions.rb
db/migrate/20200403104228_add_code_data_to_submission.rb

Don't forget to update schema version after this operation.

app/controllers/submissions_controller.rb
```diff
@@ -211,16 +227,39 @@ def submission_params
...
+    @s3_direct_post = S3_BUCKET.presigned_post(key: get_s3_key, success_action_status: '201', acl: 'private')
+  end
+
+  def handle_code_based_submissions
+    return if params[:submission][:submission_type] != "code"
+    file_location = get_s3_key.gsub("${filename}", "editor_input." + params[:language])
+    S3_BUCKET.object(file_location).put(body: params[:submission][:submission_data])
+
+    params[:submission].merge!({
+      :submission_files_attributes => {"0" => {:seq => "", :submission_type => params[:language], :submission_file_s3_key => file_location}}
```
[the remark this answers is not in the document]
skbly7: I tried to use the new hash syntax, but "0" is creating trouble for me. Do you know how that will be possible? I tried `{:0 => {:sea [...]` which gave error ultimately. 😞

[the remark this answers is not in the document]
piotrekpasciak: `'0':` should work for you.

app/services/evaluations_api_service.rb
skbly7: We are unaware of endpoint/feature changes & this feature is beta, so will need to skip test case right now.

app/controllers/submissions_controller.rb
```diff
@@ -211,16 +227,39 @@ def submission_params
...
-                            key:                   "submission_files/challenge_#{@challenge.id}/#{SecureRandom.uuid}_${filename}",
-                            success_action_status: '201',
-                            acl:                   'private')
+    @s3_direct_post = S3_BUCKET.presigned_post(key: get_s3_key, success_action_status: '201', acl: 'private')
+  end
+
+  def handle_code_based_submissions
+    return if params[:submission][:submission_type] != "code"
+    file_location = get_s3_key.gsub("${filename}", "editor_input." + params[:language])
+    S3_BUCKET.object(file_location).put(body: params[:submission][:submission_data])
```
piotrekpasciak, line 246: This looks like a request to external API. Does `S3_BUCKET` handle errors e.g network errors, validations?  

Should `create` action execute in case of error in this piece of code?

app/controllers/submissions_controller.rb
```diff
@@ -211,16 +227,39 @@ def submission_params
...
+
+    params[:submission].merge!({
+      submission_files_attributes: {"0": {seq: "", submission_type: params[:language], submission_file_s3_key: file_location}}
+    })
+  end
+
+  def handle_artifact_based_submissions
+    return if !params[:submission][:submission_files_attributes].has_key?("0")
+    return if params[:submission][:submission_files_attributes]["0"][:submission_type] != "artifact"
```
piotrekpasciak, line 255: You should be able to replace these 2 conditions with. something like:
`params.dig('submission', 'submission_files_attributes', '0', 'submission_type') != 'artifact'

app/controllers/submissions_controller.rb
```diff
@@ -211,16 +227,39 @@ def submission_params
...
+    # TODO: Make this accepted extension list dynamic via evaluations API
+    accepted_formats = [".csv", ".ipynb", ".pt"]
```
piotrekpasciak, line 257: A preferred approach is to define a constant on top of class e.g ARTIFACT_SUBMISSIONS_ACCEPTED_FORMATS = [".csv", ".ipynb", ".pt"].freeze

app/controllers/submissions_controller.rb
```diff
@@ -9,6 +9,9 @@ class SubmissionsController < ApplicationController
   before_action :set_s3_direct_post, only: [:new, :edit, :create, :update]
   before_action :set_submissions_remaining, except: [:show, :index]
   before_action :set_current_round, only: :index
+  before_action :set_form_type, only: :new
+  before_action :handle_code_based_submissions, only: [:create]
```
piotrekpasciak, line 13: That's a lot of custom business logic. A preferred approach is to move it to separate class e.g Service and write test coverage.

app/javascript/stimulus_js/controllers/code_based_submission_controller.js
```diff
@@ -0,0 +1,33 @@
+import { Controller } from 'stimulus';
+    
+export default class extends Controller {
+  connect() {
+    if (localStorage.getItem("aceEditorTheme") === null) {
+      localStorage.setItem("aceEditorTheme", "ace/theme/github");
+    }
+    
+    $(document).ready(function() {
+      var submission = ace.edit('ace-editor');
```
piotrekpasciak, line 10: You should avoid using `var` keyword 
- because var variables can be redeclared what can generate nasty bugs
- var variables have global / function scope instead of expected block scope
- using var is considered legacy code approach

Please use `const` instead

https://dev.to/sarah_chima/var-let-and-const--whats-the-difference-69e

app/jobs/submission_grader_job.rb
```diff
@@ -1,8 +1,16 @@
 class SubmissionGraderJob < ApplicationJob
   queue_as :default
 
+  def get_evaluator_type(submission_id)
```
piotrekpasciak, line 4: Shouldn't that be a private method?

app/jobs/submission_grader_job.rb
```diff
@@ -1,8 +1,16 @@
...
+    challenge = Submission.find(id=submission_id).challenge
```
piotrekpasciak, line 5: `id=submission_id` looks. weird. Don't you mean simply `Submission.find(submission_id)`?

app/controllers/submissions_controller.rb
piotrekpasciak: You can definitely return a value from that class and assign it to variable.

app/controllers/submissions_controller.rb
```diff
@@ -211,16 +227,43 @@ def submission_params
...
+      redirect_to challenge_submissions_path(@challenge), notice: 'Submission upload failed, please try again.'
+    end
+
+    params[:submission].merge!({
+      submission_files_attributes: {"0": {seq: "", submission_type: params[:language], submission_file_s3_key: file_location}}
+    })
+  end
+
+  def handle_artifact_based_submissions
+    return if !params.dig('submission', 'submission_files_attributes').has_key?("0")
```
piotrekpasciak, line 258: You can get rid of this `return if` now, because, second one already checks this scenario

PR #1364, opened 2020-04-20
Return masthead partial

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/522766588
Expose challenge Masterhead as API

### DIFF
--- a/app/controllers/api/v1/challenges/challenges_controller.rb
+++ b/app/controllers/api/v1/challenges/challenges_controller.rb
@@ -58,8 +58,8 @@ module Api
         private
 
         def set_challenge
-          @challenge = Challenge.friendly.find(params[:id])
-          authorize @challenge, :edit?
+          @challenge = Challenge.find_by_slug(params[:id])
+          authorize @challenge, :edit? if @challenge
         end
 
         def set_challenge_to_discource
--- a/app/controllers/api/v1/challenges/challenges_controller.rb
+++ b/app/controllers/api/v1/challenges/challenges_controller.rb
@@ -55,7 +55,7 @@ module Api
 
         def set_challenge
           @challenge = Challenge.friendly.find(params[:id])
-          authorize @challenge, :edit? if @challenge
+          authorize @challenge, :edit?
         end
 
         def set_challenge_to_discource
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -30,12 +39,44 @@ def update
           end
         end
 
+        def masthead
```
piotrekpasciak, line 42: We could use some basic specs for this endpoint.

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -30,12 +39,44 @@ def update
...
+          if @challenge.present?
```
piotrekpasciak, line 43: This if statement looks to be redundant and you're never going to reach else condition.

Isn't `Challenge.friendly.find(params[:id])` going to raise 404 error when challenge not presennt?

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -30,11 +39,43 @@ def update
...
+            render json: { error: 'Challenge not present.' }, status: :unprocessable_entity
+          end
+        end
+
         private
 
         def set_challenge
-          @challenge = Challenge.friendly.find(params[:id])
-          authorize @challenge, :edit?
+          @challenge = Challenge.find_by_slug(params[:id])
```
piotrekpasciak, line 61: Previous version was mutch better: 
`Challenge.friendly.find(params[:id])`

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -30,11 +39,43 @@ def update
...
+            render json: {
+              masthead:
+                render_to_string(
+                  partial: 'shared/challenges/masthead',
+                  formats: :html,
+                  layout: false,
+                  locals: { challenge: @challenge, challenge_rounds: @challenge_rounds, vote: @vote, follow: @follow }
+                )
+            }
+          else
```
piotrekpasciak, line 53: Get rid of this `if else` block after you'll use 
`Challenge.friendly.find(params[:id])`

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -2,8 +2,17 @@ module Api
   module V1
     module Challenges
       class ChallengesController < ::Api::V1::BaseController
-        before_action :auth_by_participant_api_key, only: [:create, :update]
-        before_action :set_challenge, only: :update
+        include ActionController::Helpers
```
piotrekpasciak, line 5: What kind of `ActionController::Helpers` do we need here?

alkeshrails: partial `shared/challenges/masthead` need, below listed helper ChallengesHelper, VotesHelper, FollowsHelper. so use that helper I need to `include ActionController::Helpers`

piotrekpasciak: Alright, I thought including only our helpers would be sufficient. Looks good then.

app/controllers/api/v1/challenges/challenges_controller.rb
```diff
@@ -30,11 +39,39 @@ def update
...
+              )
+          }
+        end
+
         private
 
         def set_challenge
           @challenge = Challenge.friendly.find(params[:id])
-          authorize @challenge, :edit?
+          authorize @challenge, :edit? if @challenge
```
piotrekpasciak, line 58: We should get rid of `if @challenge` here and setters below.

PR #1366, opened 2020-04-20
Remove organizer_id from participants

alkeshrails:
and fix related issues


db/seeds.rb
```diff
@@ -698,8 +696,8 @@
     email: "p7@example.com",
     name: "p7")
 
-  ParticipationTerms.create!(
-    terms_markdown: "Sample Participation Terms. Please update at '/admin/participation_terms'"
-  )
+  # ParticipationTerms.create!(
```
piotrekpasciak, line 699: Are you sure you should remove ParticipationTems? What's the issue with it?

db/seeds.rb
```diff
@@ -97,8 +95,8 @@
 
   # Partners
   Partner.create!([
-    {id: 22, organizer_id: 21 , image_file: open('https://dnczkxd1gcfu5.cloudfront.net/images/challenges/image_file/50/spotify.png'), name: "Spotify", visible: true},
-    {id: 1223, organizer_id: 17 , image_file: open('https://dnczkxd1gcfu5.cloudfront.net/images/challenges/image_file/50/spotify.png'), name: "Uber", visible: true}
+    {id: 22, image_file: open('https://dnczkxd1gcfu5.cloudfront.net/images/challenges/image_file/50/spotify.png'), name: "Spotify", visible: true},
```
piotrekpasciak, line 98: These are partners. Shouldn't they keep organizer_id?

db/seeds.rb
alkeshrails: Because terms_markdown column is removed from participation_terms.

piotrekpasciak: There is still `terms` column. Please use it to not remove records from seed file.

alkeshrails: okay, thanks

db/schema.rb
```diff
@@ -11,7 +11,6 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema.define(version: 2020_04_19_194412) do
```
piotrekpasciak, line 13: Please fix the schema.rb to:
- include correct timestamp
- include changes only from your migrations

PR #1367, opened 2020-04-21
Render sponsor logos in challenge overview

alkeshrails:
image


app/models/organizer.rb
```diff
@@ -10,6 +10,7 @@ class Organizer < ApplicationRecord
   has_many :participants, dependent: :nullify, through: :participant_organizers
   has_many :clef_tasks, dependent: :destroy
   has_many :challenge_calls, dependent: :destroy
+  has_many :partners
```
piotrekpasciak, line 13: You should specify `dependent` option here. Probably `:nullify` is right. one.

PR #1368, opened 2020-04-21
Fix Discourse API request limit issue

piotrekpasciak:


app/models/challenge.rb
```diff
@@ -101,6 +101,7 @@ def create_discourse_category
 
   def update_discourse_category
     return if Rails.env.development? || Rails.env.test?
+    return unless saved_change_to_attribute?(:challenge)
```
skbly7, line 104: We were changing `featured_sequence` in case of reorder - which caused the issue, will this function work for that?

piotrekpasciak: Request will be sent only when `challenge` field. was changed (title challenge).

Seq field change should not trigger request after this change.

skbly7: oh, perfect!! I misunderstood ":challenge" as Challenge model instead of `challenge` field.

PR #1369, opened 2020-04-21
Add edit link to admin challenge index

alkeshrails:


app/admin/challenge.rb
```diff
@@ -27,11 +31,19 @@
     column :page_views
     column :participant_count
     column :submissions_count
-    actions
+    column '', class: 'col-actions'do |resource|
+      div class: 'table_actions' do
```
piotrekpasciak, line 35: I also had problem with rendering multiple links in ActiveAdmin and I've resolved it this way:
https://github.com/AIcrowd/AIcrowd/blob/master/app/admin/newsletter_email.rb#L44

I think it would be cleaner to use `a` method instead of this approach.

alkeshrails: @piotrekpasciak good stuff, have cleaned up, please review again.

PR #1372, opened 2020-04-22
An endpoint for getting user details from AIcrowd API Key

YoogottamK:


app/controllers/api/base_controller.rb
```diff
@@ -16,6 +18,25 @@ def auth_by_api_key
     end
   end
 
+
+  def user_auth_by_api_key
+    authenticate_or_request_with_http_token do |token, options|
+      user = Participant.find_by(api_key: token)
+
+      return nil if not user.present?
```
skbly7, line 26: Can we make it more informative in case of an invalid API key? Also play around with status code like 200, 401, etc.

app/controllers/api/base_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'json/ext'
```
piotrekpasciak, line 1: Why do you add this require?

app/controllers/api/base_controller.rb
```diff
@@ -16,6 +18,25 @@ def auth_by_api_key
     end
   end
 
+
+  def user_auth_by_api_key
```
piotrekpasciak, line 22: This is. abstract class,  it's not meant to define new action here. It's violation of basic object programming rules.

Please define new action in proper controller like `participants_controlller` under `api/v1` namespace.

app/controllers/api/base_controller.rb
```diff
@@ -16,6 +18,25 @@ def auth_by_api_key
...
+    authenticate_or_request_with_http_token do |token, options|
+      user = Participant.find_by(api_key: token)
+
+      return nil if not user.present?
+
+      render json: {
+        "username" => user.name,
```
piotrekpasciak, line 29: Please use. a Serializer class to return response. You may need to define new one for participants.

config/routes.rb
```diff
@@ -74,6 +74,7 @@
     get 'user', to: 'oauth_credentials#show'
     get 'mailchimps/webhook' => 'mailchimps#verify', :as => :verify_webhook
     post 'mailchimps/webhook' => 'mailchimps#webhook', :as => :update_webhook
+    post 'auth/api_key', to: 'base#user_auth_by_api_key'
```
piotrekpasciak, line 77: Please move it under `api/v1` namespace and use `routes`. helpers.

piotrekpasciak:
@YoogottamK I've updated the PR and endpoint is under `/api/v1/api_user.json` path. Merge the PR whenever you need it.

PR #1376, opened 2020-04-22
Fix Ui of CKeditor for admin also fix ul element

alkeshrails:
Ck Editor Ui fix and fix for `ul` `ol` list
Screenshot from 2020-04-22 15-49-38

### DIFF
--- a/app/admin/job_posting.rb
+++ b/app/admin/job_posting.rb
@@ -46,7 +46,8 @@ ActiveAdmin.register JobPosting do
       f.input :location
       f.input :country
       f.input :remote
-      f.input :description, as: :ckeditor
+      f.label 'Description', class: 'ckeditor_label job_posting'
+      f.text_area :description, class: 'ckeditor'
       f.input :job_url
     end
     f.actions
--- a/app/assets/stylesheets/ckeditor.scss
+++ b/app/assets/stylesheets/ckeditor.scss
@@ -1,8 +1,21 @@
 .cke_chrome {
-  width: 50% !important;
+  width: 75% !important;
   overflow: hidden;
 }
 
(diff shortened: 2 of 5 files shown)


app/admin/job_posting.rb
```diff
@@ -46,7 +46,8 @@ def permitted_params
       f.input :location
       f.input :country
       f.input :remote
-      f.input :description, as: :ckeditor
+      f.label 'Description', class: 'ckeditor_label job_posting'
```
skbly7, line 49: Why  it should be different class names everywhere? Instead of just one say `admin-panel` etc.

alkeshrails: Actually spacing is diff on pages, so I need to set margin accordingly.

app/assets/stylesheets/ckeditor.scss
```diff
@@ -1,8 +1,21 @@
...
+  width: 75% !important;
   overflow: hidden;
 }
 
 .cke_1.cke.cke_reset.cke_chrome.cke_ltr.cke_browser_webkit {
   margin-left: 25%;
+  margin-bottom: 2%;
+}
+
+label.ckeditor_label.job_posting, label.ckeditor_label.participation_terms {
```
piotrekpasciak, line 11: Are you sure you want to modify all CKEditor in AICrowd?

If you want to modify css for Admin panel CKEditor use following stylesheet:
https://github.com/AIcrowd/AIcrowd/blob/master/app/assets/stylesheets/additional_active_admin.css.scss

piotrekpasciak: @alkeshrails said this ckeditor.scss is used in Admin panel and front-end is usinng different css file.

PR #1380, opened 2020-04-23
fix for google analytics to track hash

nikhilrayaprolu:


nikhilrayaprolu:
Using https://support.google.com/analytics/thread/16933523?hl=en and added the respective lines.

PR #1381, opened 2020-04-24
get user profile tooltip to API call

alkeshrails:
Make the tooltips of users come via API instead of pre-rendered!
Monday : https://aicrowd.monday.com/boards/363093650/pulses/522768696

User tooltip
Screenshot from 2020-04-24 13-25-05

### DIFF
--- a/app/javascript/stimulus_js/controllers/participants_controller.js
+++ b/app/javascript/stimulus_js/controllers/participants_controller.js
@@ -1,0 +1,23 @@
+import { Controller } from 'stimulus';
+export default class extends Controller {
+
+  getUserProfile(event){
+    const avatar        = this.data.get('avatar');
+    const location      = this.data.get('location');
+    const username      = this.data.get('username');
+    const participantId = this.data.get('participant');
+
+    $.ajax({
+      url: '/api/v1/participants/'+ participantId +'/user_profile',
+      type: 'GET',
+      data: {avatar: avatar, username: username},
+      success:  function(result){
+                  jQuery("#"+location).html(result);
+                  $('[data-toggle="tooltip"]').tooltip();
+                  $('[data-toggle="popover"]').popover();
+                  hljs.initHighlightingOnLoad();
+                }
+    })
+    console.log("Get user #" + participantId + " detail.");
+  }
+}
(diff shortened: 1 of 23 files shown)


app/javascript/stimulus_js/controllers/participants_controller.js
```diff
@@ -0,0 +1,23 @@
+import { Controller } from 'stimulus';
+export default class extends Controller {
+
+  getUserProfile(event){
+    const avatar        = this.data.get('avatar');
+    const location      = this.data.get('location');
+    const username      = this.data.get('username');
+    const participantId = this.data.get('participant');
+
+    $.ajax({
```
piotrekpasciak, line 10: We should avoid jQuery whenever possible as it's legacy library. Please use vanilla javascript method e.g `fetch` or `async await`:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await

app/views/participants/_user_link.html.erb
```diff
@@ -1,17 +1,14 @@
-<% 
-    uniq_id = rand(1000)
-    random_class = 'tooltip-user-wrapper-' + uniq_id.to_s + '-' + participant.id.to_s
+<% avatar = avatar.present? ? avatar : true %>
+<% username = username.present? ? username : true %>
+<%
+  uniq_id = rand(1000)
```
piotrekpasciak, line 4: SecureRandom.hex()

app/views/participants/_name_avatar.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if !defined?(avatar) || (defined?(avatar) && avatar && avatar != 'false') %>
```
piotrekpasciak, line 1: `(defined?(avatar) && avatar != 'false')`  will be simpler and work the same way.

app/views/participants/_name_avatar.html.erb
```diff
@@ -0,0 +1,6 @@
...
+  <%= participant_avatar participant %>
+<% end %>
+<% if !defined?(username) || (defined?(username) && username && username != 'false') %>
```
piotrekpasciak, line 4: (defined?(username) && username != 'false')

app/javascript/stimulus_js/controllers/participants_controller.js
alkeshrails: By using `fetch` I didn't get `render partial` of tooltip in response, that we need to set as `html` .

PR #1384, opened 2020-04-25
enable link in flash messages

alkeshrails:


app/views/layouts/_flash_message.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="alert alert-<%= rewrite_type type %> alert-dismissible fade show flash-message" role="alert">
   <div class="container-fluid">
     <center>
-      <%= message %>
+      <%= sanitize(message) %>
```
piotrekpasciak, line 4: Why did you choose to use `sanitize` instead of `sanitize_html`?

alkeshrails: @piotrekpasciak I have applied this method since being used throughout the application. please review.

PR #1389, opened 2020-04-27
Fix status filter JS and nil, blank prize filter

alkeshrails:
Bug/ staging - On challenge filter not working properly
Screenshot from 2020-04-27 14-00-38

### DIFF
--- a/app/javascript/stimulus_js/controllers/challenges_controller.js
+++ b/app/javascript/stimulus_js/controllers/challenges_controller.js
@@ -21,7 +21,7 @@ export default class extends Controller {
     let url = window.location.origin + window.location.pathname
     let params = ''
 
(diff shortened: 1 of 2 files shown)


app/javascript/stimulus_js/controllers/challenges_controller.js
```diff
@@ -21,7 +21,7 @@ export default class extends Controller {
     let url = window.location.origin + window.location.pathname
     let params = ''
 
-    if(status.value !== undefined){
+    if(status !== null && status.value !== undefined){
```
piotrekpasciak, line 24: In javascript, when you want to check if string is present or collectin has elements you use `status.length`. It's more clean and reliable then your check. 

In javascript 0 evaluates to false in contrary to Ruby.

PR #1394, opened 2020-04-29
set footer to DB column

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/547029065

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -11,93 +11,97 @@
 <% end %>
 
 <!-- app footer -->
-<footer class="app-footer">
-	<div class="container-fluid">
-		<div class="row">
+<% if footer_body.present? %>
+	<%= footer_body %>
+<% else %>
+	<footer class="app-footer">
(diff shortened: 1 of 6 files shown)


db/migrate/20200429101948_create_footers.rb
```diff
@@ -0,0 +1,9 @@
+class CreateFooters < ActiveRecord::Migration[5.2]
```
piotrekpasciak, line 1: I think that creating separate table. for. footer is an overkill. Please reuse `settings` table from previous PR.

app/views/layouts/_footer.html.erb
```diff
@@ -11,93 +11,97 @@
 <% end %>
 
 <!-- app footer -->
-<footer class="app-footer">
-	<div class="container-fluid">
-		<div class="row">
+<% if footer_body.present? %>
+	<%= footer_body %>
```
piotrekpasciak, line 15: What will this footer contain?  If mostly. javascript code it should be added in the end of document because. of performance reasons.

PR #1411, opened 2020-05-01
intercom integration

nikhilrayaprolu:


app/services/intercom_service.rb
```diff
@@ -0,0 +1,15 @@
+class IntercomService
+  def initialize
+    @client = Intercom::Client.new(token: ENV['INTERCOM_TOKEN'])
+  end
+
+  def send_event(event_name, participant, metadata = {})
```
piotrekpasciak, line 6: Our service classes should implement only `initialize` and `call` public methods. Let's stick to this convention.

piotrekpasciak, line 6: Are you sure this class won't fail? Usually, it's needed to handle network errors and endpoint validations when sending request to external service.

app/services/intercom_service.rb
```diff
@@ -0,0 +1,15 @@
+class IntercomService
```
piotrekpasciak, line 1: Please add test coverage for this class. You can use VCR for that.

spec/services/intercom_service_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'rails_helper'
```
nikhilrayaprolu, line 1: @piotrekpasciak 
This is an incomplete test, but need a lot of input from you.
1) I am not able to make the described_class use environment variables - How many types I tried, I get
`IntercomService#initialize
Failure/Error: @client = Intercom::Client.new(token: ENV['INTERCOM_TOKEN'])
     
     Intercom::MisconfiguredClientError:
       an access token must be provided
`
you can see that inside my class I have
`
def initialize
    begin
      @client = Intercom::Client.new(token: ENV['INTERCOM_TOKEN'])
    rescue Intercom::MisconfiguredClientError => e
      Rails.logger.info "Intercom: #{e.message}"
      raise
    end
`

2) How can I use VCR here?, I don't have any URL being called explicitly, intercom-rails library handles all the API calls.

3) Also, can you just go through the tests though those are too naive and not yet the right tests, tell me whether they capture of the essence of testing.

piotrekpasciak: 1. Probably it's because of following line:
https://github.com/AIcrowd/AIcrowd/pull/1411/files#diff-5ecffe20d437d5d1b021927bea980c1fR19

In line above you're removing ENV variable for every test. You should set value for single test via stubbing. Consider this example: 
before { ENV.stub(:[]).with('INTERCOM_TOKEN').and_return('') }``

piotrekpasciak: 2.  Take a look on following example:
https://github.com/AIcrowd/AIcrowd/blob/master/spec/services/discourse/add_users_to_group_service_spec.rb#L28

You simply wrap the piece of code that sends request to external service.

piotrekpasciak:
@nikhilrayaprolu I've added commit that updates IntercomService that it would basic Object Oriented Programming rules, as it would be way to mutch comments in Github to fix it.

app/models/submission.rb
```diff
@@ -51,6 +51,7 @@ class Submission < ApplicationRecord
         .perform_later(challenge_round_id: challenge_round_id)
     end
     Prometheus::SubmissionCounterService.new(submission_id: id).call
+    IntercomService.new.send_event('made_submission', Participant.find_by(id: participant_id))
```
piotrekpasciak, line 54: Are you sure this works? You've `call` method inside IntercomService. I haven't seen `send_event` method there.

spec/services/intercom_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    context 'with a right event name' do
+      it 'returns null with right event name' do
+        VCR.use_cassette('intercom_service/with_right_event_name') do
+          expect(IntercomService.new('test_event', participant, {}).call).to eq nil
+        end
+      end
+    end
+
+    context 'with no participant' do
+      it 'returns raises excetion of Intercom' do
```
piotrekpasciak, line 32: Typo, should be `exception`

spec/services/intercom_service_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'rails_helper'
+
+describe IntercomService do
+  let!(:participant) { create :participant }
+
+  describe '#call' do
+    context 'with no event name'  do
+      it 'returns raises excetion of Intercom' do
```
piotrekpasciak, line 8: Typo, should be `exception`.

spec/services/intercom_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    context 'with no participant' do
+      it 'returns raises excetion of Intercom' do
+        VCR.use_cassette('intercom_service/with_no_participant') do
+          expect(IntercomService.new('test_event', nil, {}).call).to eq nil
+        end
+      end
+    end
+
+    context 'with a participant not having email' do
+      it 'returns raises excetion of Intercom' do
```
piotrekpasciak, line 40: exception

spec/services/intercom_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        VCR.use_cassette('intercom_service/with_participant_not_having_email') do
+          expect(IntercomService.new('test_event', participant, {}).call).to eq nil
+        end
+      end
+    end
+
+    context 'with metadata has a hash' do
+      it 'returns raises excetion of Intercom' do
```
piotrekpasciak, line 48: exception

PR #1422, opened 2020-05-04
Add dataset folders functionality

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/546476740
https://aicrowd.monday.com/boards/363093650/pulses/546487031

Screenshot 2020-05-11 at 00 44 39
Screenshot 2020-05-11 at 00 46 24

### DIFF
--- a/db/migrate/20200510182517_remove_own_fields_from_dataset_files.rb
+++ b/db/migrate/20200510182517_remove_own_fields_from_dataset_files.rb
@@ -1,0 +1,9 @@
+class RemoveOwnFieldsFromDatasetFiles < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :dataset_files, :directory_path, :text
+    remove_column :dataset_files, :aws_access_key, :text
+    remove_column :dataset_files, :aws_secret_key, :text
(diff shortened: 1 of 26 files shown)


db/migrate/20200510182517_remove_own_fields_from_dataset_files.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveOwnFieldsFromDatasetFiles < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :dataset_files, :directory_path, :text
+    remove_column :dataset_files, :aws_access_key, :text
+    remove_column :dataset_files, :aws_secret_key, :text
+    remove_column :dataset_files, :bucket_name, :text
+    remove_column :dataset_files, :region, :text
```
alkeshrails, line 7: This column is not present in below schema.
dataset_folder has this column.

piotrekpasciak: It's because 2 previous migrations had impact on schema.rb. When you analyze all 3 migrations you'll notice that column is not meant to be in schema.rb.

PR #1424, opened 2020-05-04
Change banner_text type to admin

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/548562076

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -139,7 +139,7 @@ module ApplicationHelper
   end
 
   def banner_text
-    Setting.banner_record&.gsub('<p>', '')&.gsub('</p>', '')&.html_safe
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -139,14 +139,14 @@ def social_image_url(site)
...
+    sanitize_html Setting.banner_record&.gsub('<p>', '')&.gsub('</p>', '')
   end
 
   def banner_color
     Setting.banner_color_value
   end
 
   def footer_text
-    Setting.footer_record
+    sanitize_html Setting.footer_record
```
piotrekpasciak, line 150: Wasn't we supposed to allow any kind of javascript and HTML code here? To achieve it we probably have to use `raw` method instead.

PR #1425, opened 2020-05-04
Implement infinite scroll

alkeshrails:
instead of load more button


app/javascript/stimulus_js/controllers/challenges_controller.js
```diff
@@ -1,4 +1,6 @@
 import { Controller } from 'stimulus';
+import { load_more_data } from '../helpers/load_more_data_helper'
```
piotrekpasciak, line 2: I don't see file load_more_data_helper.js here. Are you sure it exists?

alkeshrails: ahh, not sure how I missed to remove it.

skbly7:
As shared in the team standup today, everything looks good (haven't checked code), only thing which needs to happen is move away from full page loader to showing loader only at the bottom of page.

PR #1429, opened 2020-05-05
Add calendar heatmap to user profile

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/551540344

Screenshot 2020-05-07 at 00 07 02

Link to review app:
https://aicrowd-feature-add-act-eme0vg.herokuapp.com/participants/mtab


skbly7:
Hey, the preview looks beautiful.

Without going to the code I have following suggestions:

1. Visit-1 & Submission-10, can be moved to right instead of left (or move 1-10, 10-.... thing to right)
2. Use shades of AIcrowd red (#F0524D), instead of brown.

skbly7:
Also feel free to move it at top! 😄 i.e. above the line chart + pie chart

piotrekpasciak:
@skbly7 I've applied remarks mentioned by you. Take a look on screenshot.

skbly7:
Thanks, that looks great. I will dive into the code.

app/queries/participants/activity_query.rb
```diff
@@ -0,0 +1,22 @@
...
+  class ActivityQuery
+    VALUE_OF_VISIT      = 1.freeze
+    VALUE_OF_SUBMISSION = 10.freeze
+
+    def initialize(participant:)
+      @participant = participant
+    end
+
+    def call
+      visits      = participant.visits.group_by_day(:started_at, range: 1.year.ago.midnight..Time.current).count
```
skbly7, line 11: Side Note: Do check out how much visits retention we have, I think we keep only 30 days of visit data. 
Changing it to higher number may not be best way either, lets have a batch job which can convert number of visits data into a new table daily.

piotrekpasciak: Screenshot 2020-05-07 at 10 28 33

As far I know we have full retention on visits. Ahoy::Visit was added to codebase 7 months
ago and last records from that time are still in database.

piotrekpasciak:
Feedback from Mohanty:
`But in general, looks great. If we can solve the mobile view issue that's great, but it shouldn't be a blocker from rolling this out to prod`

PR #1431, opened 2020-05-05
adding weights to the challenges

nikhilrayaprolu:


db/migrate/20200505093714_add_weight_to_challenge.rb
```diff
@@ -0,0 +1,5 @@
+class AddWeightToChallenge < ActiveRecord::Migration[5.2]
+  def change
+    add_column :challenges, :weight, :integer, default: 0
```
piotrekpasciak, line 3: Generally, it's good idea to add `null: false` together with `default` option.

nikhilrayaprolu: made a new migration reflecting the same

alkeshrails: @nikhilrayaprolu This is open PR, so you can rollback migration and create a single migration with column type: float, default 0 and null false

alkeshrails: Please pull schema.rb as well

PR #1433, opened 2020-05-06
 Show number of team statistic

alkeshrails:
in challenge masthead
image

### DIFF
--- a/app/views/shared/challenges/_listing.html.erb
+++ b/app/views/shared/challenges/_listing.html.erb
@@ -78,7 +78,7 @@
               <%= challenge_stat_count(challenge, 'participant') %>
             </span>
           </div>
-
+          <%= render partial: 'shared/challenges/team_stat', locals: { challenge: challenge } %>
           <div class="stat-item ml-1 mr-1">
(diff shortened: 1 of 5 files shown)


app/views/shared/challenges/_masthead_extra.html.erb
```diff
@@ -30,6 +30,14 @@
         </span>
         </center>
       </div>
+      <div class="stat-item" style="width: 72px">
+        <center>
```
skbly7, line 34: Show it in other places wherever stats are being shown. I think you will need to make change in `shared/challenge/_listing` or so for that.

app/views/shared/challenges/_listing.html.erb
```diff
@@ -78,7 +78,14 @@
               <%= challenge_stat_count(challenge, 'participant') %>
             </span>
           </div>
-
+          <div class="stat-item" style="width: 72px">
```
piotrekpasciak, line 81: This should be moved to partial in `shared/` directory. So we can modify it in one place.

PR #1435, opened 2020-05-06
Redirect after participants accepts terms to previous page

alkeshrails:
instead of challenge homepage


app/controllers/challenge_participants_controller.rb
```diff
@@ -22,7 +22,7 @@ def update
     @challenge_participant.challenge_rules_accepted_date    = Time.now
     @challenge_participant.challenge_rules_accepted_version = @challenge_participant.challenge.current_challenge_rules&.version
     if @challenge_participant.update(challenge_participant_params)
-      redirect_to(session[:forwarding_url] || @challenge_participant.challenge)
+      redirect_to(stored_location_for(:user) || session[:forwarding_url])
```
skbly7, line 25: Both the variables i.e. `stored_location_for` & `session[:forwarding_url]` can technically be not present on particular request. Hence, we should leave a default value which will 100% work like `@challenge_participant.challenge`.

Issue #1204, opened 2020-03-06
Changing Participant username changes Discourse username only after logout


piotrekpasciak:

skbly7:
Change username -> Callback to Discourse for syncing information.

PR #1438, opened 2020-05-07
Discource service for update username

alkeshrails:
Issue #1204: Synchronize participant username with Discourse
Monday : https://aicrowd.monday.com/boards/363093650/pulses/546451555


piotrekpasciak:
This is no longer needed

PR #1439, opened 2020-05-07
Set default 1 for challenge featured_sequence

alkeshrails:
Features Challenge ordering should not be affected by new challenge creations (etc)
Monday : https://aicrowd.monday.com/boards/363093650/pulses/546915993

Run `rake update_featured_sequence:increase_by_one`

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -70,6 +70,7 @@ class Challenge < ApplicationRecord
             format: { with: /\A[a-zA-Z0-9]/ }
   validates :challenge_client_name, presence: true
   validate :other_scores_fieldnames_max
+  validate :greater_than_zero
   #validate :banner_color, format: { with: /\A#?(?:[A-F0-9]{3}){1,2}\z/i }
 
   EVALUATOR_TYPES = {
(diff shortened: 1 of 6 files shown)


app/models/challenge.rb
```diff
@@ -194,6 +195,10 @@ def other_scores_fieldnames_max
     errors.add(:other_scores_fieldnames, 'A max of 5 other scores Fieldnames are allowed') if other_scores_fieldnames && (other_scores_fieldnames.count(',') > 4)
   end
 
+  def greater_than_zero
+    errors.add(:featured_sequence, 'should be greater than zero') if featured_sequence <= 0
```
piotrekpasciak, line 199: Add `featured_sequence.to_i` so we won't run into nil exception here.

lib/tasks/update_featured_sequence.rake
```diff
@@ -0,0 +1,11 @@
+namespace :update_featured_sequence do
+  desc "set featured_sequence"
+  task increase_by_one: :environment do
+    challenges = Challenge.where(featured_sequence: 0)
+    challenges.each do |challenge|
+      sleep 3
```
piotrekpasciak, line 6: Why are you using sleep here?

alkeshrails: At my local, I have many challenge with featured_sequence value 0
I ran this rake and some number is skip.
like after 20, 21 I get next featured_sequence 23.
so that I add sleep.

alkeshrails:
Issue : 
Screenshot from 2020-05-08 13-24-19

After rake task run 
Screenshot from 2020-05-08 14-12-27

lib/tasks/update_featured_sequence.rake
piotrekpasciak: But sleep doesn't skip anything. It only makes script running slower.

piotrekpasciak: I feel like you could simply remove it.

PR #1444, opened 2020-05-08
Truncate team name

alkeshrails:
Truncate Team name

### DIFF
--- a/app/views/challenges/show/_top_five_leaderboards.html.erb
+++ b/app/views/challenges/show/_top_five_leaderboards.html.erb
@@ -1,3 +1,4 @@
+<% truncate = truncate.present? ? truncate : false %>
 <header class="section-header-decorative mt-4">
   <div>
     <h4>Leaderboard</h4>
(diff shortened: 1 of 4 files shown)


app/views/challenges/show/_top_five_leaderboards.html.erb
```diff
@@ -1,3 +1,4 @@
+<% truncate = truncate.present? ? truncate : false %>
```
piotrekpasciak, line 1: I think you mean more something like `truncate.defined?`.

app/views/leaderboards/_leaderboards.html.erb
```diff
@@ -2,6 +2,7 @@
   <%
     participants = leaderboard_participants(leaderboard)
     challenge    = leaderboard.challenge
+    truncate     = truncate.present? ? truncate : false
```
piotrekpasciak, line 5: `truncate.defined?`

alkeshrails: undefined method `defined?' for "true":String

app/views/challenges/show/_top_five_leaderboards.html.erb
alkeshrails: I get error undefine method ``defined?`` for nil or string.

alkeshrails:
undefined method `defined?' for "true":String

PR #1447, opened 2020-05-08
Feature: Leaderboard recomputations on challenge update

skbly7:


app/controllers/challenges_controller.rb
```diff
@@ -81,13 +81,20 @@ def create
     end
   end
 
+  def leaderboard_recomputations
```
piotrekpasciak, line 84: Please move it to private section.

skbly7: Moved, thanks.

PR #1450, opened 2020-05-11
Fix space after teams count

alkeshrails:
image

image

### DIFF
--- a/app/views/shared/challenges/_listing.html.erb
+++ b/app/views/shared/challenges/_listing.html.erb
@@ -78,14 +78,7 @@
               <%= challenge_stat_count(challenge, 'participant') %>
             </span>
           </div>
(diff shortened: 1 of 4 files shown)


app/views/shared/challenges/_listing.html.erb
```diff
@@ -78,7 +78,14 @@
               <%= challenge_stat_count(challenge, 'participant') %>
             </span>
           </div>
-          <%= render partial: 'shared/challenges/team_stat', locals: { challenge: challenge } %>
```
piotrekpasciak, line 81: Please don't remove this shared partial. Instead pass to it `stat_item_class` variable and used it like `<div class="stat-item <%= stat_tiem-class if defined?(stat_item_class)%>">`

PR #1464, opened 2020-05-14
Update practices route to practice

alkeshrails:
update url from /pratices to /practice

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -225,7 +225,8 @@ Rails.application.routes.draw do
   end
 
   resources :team_members, path: "our_team", only: [:index]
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -225,7 +225,8 @@ def challenge_routes
   end
 
   resources :team_members, path: "our_team", only: [:index]
-  resources :practices, only: [:index]
+
+  get '/practice', to: 'practices#index'
```
piotrekpasciak, line 229: It shoould be `resource :practices, only: [:index]`. We should use rails helpers whenever possible.

PR #1479, opened 2020-05-18
Fix prefixed hash category filter issue

alkeshrails:
1. The rake task is to remove all the existing categories with prefixed hash.
2. Command to remove all existing prefixed hash categories through the rake task is `rake category:remove_prefixed_hash_from_categories`.

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -253,12 +253,12 @@ class ChallengesController < ApplicationController
   def update_challenge_categories
     @challenge.category_challenges.destroy_all if @challenge.category_challenges.present?
     params[:challenge][:category_names].reject(&:empty?).each do |category_name|
-      begin
-        category = Category.find_or_create_by!(name: category_name)
-      rescue ActiveRecord::RecordInvalid => invalid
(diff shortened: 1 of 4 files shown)


app/controllers/challenges_controller.rb
```diff
@@ -253,8 +253,12 @@ def update_challenges_organizers
   def update_challenge_categories
     @challenge.category_challenges.destroy_all if @challenge.category_challenges.present?
     params[:challenge][:category_names].reject(&:empty?).each do |category_name|
-      category = Category.find_or_create_by!(name: category_name)
-      @challenge.category_challenges.create!(category_id: category.id)
+      begin
+        category = Category.find_or_create_by!(name: category_name)
+      rescue ActiveRecord::RecordInvalid => invalid
```
piotrekpasciak, line 258: I think there is no point to use `rescue` here. You could do it like that:
```

category = Category.find_or_create_by(name: category_name)

if category.save
  @challenge.category_challenges.create!(category_id: category.id)
else
  @challenge.errors.messages.merge!(invalid.record.errors.messages)
end
```

PR #1482, opened 2020-05-19
Fix top nav and headers

alkeshrails:


app/views/layouts/_header.html.erb
```diff
@@ -114,9 +109,6 @@
           <li class="nav-item">
```
skbly7, line 109: Remove `user rankings` too

PR #1483, opened 2020-05-19
Truncate submission message

alkeshrails:
Error display fix in submissions (trim from css and not rails side) + Show error in submission#show page
Screenshot from 2020-05-20 15-21-19

Mobile view :
AIcrowd_NeurIPS_2019_MineRL_Competition_Submissions

### DIFF
--- a/app/assets/stylesheets/application-redesign.css.scss
+++ b/app/assets/stylesheets/application-redesign.css.scss
@@ -10293,6 +10293,14 @@ button.bg-bronze:focus {
   white-space: nowrap;
 }
 
+.submission-truncate {
+  width: 579px;
+  overflow: hidden;
+  display: inline-block;
+  text-overflow: ellipsis;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application-redesign.css.scss
```diff
@@ -10293,6 +10293,14 @@ button.bg-bronze:focus {
   white-space: nowrap;
 }
 
+.submission-truncate {
```
skbly7, line 10296: Why are we fixing width? Would `width: 100%` work? (because fixing width to px value will break value in mobile I assume)

In case yes, lets have a generic `.text-trucate` class or so, instead of creating new for everyplace.

alkeshrails: Width in % is not work here.
Its change table width on page.

PR #1484, opened 2020-05-19
fixing seeds and display for badges

nikhilrayaprolu:


db/seeds/badges.rb
```diff
@@ -1,62 +1,62 @@
 AicrowdBadge.create!([
-  {name: "Trustable", description: "Has Filled his profile page", badge_type_id: 3, code: "@participant.add_badge(\"Trustable\")", badges_event_id: 10, image: nil},
+  {name: "Trustable", description: "Has Filled his profile page", badge_type_id: 3, code: "@participant.add_badge(\"Trustable\")", badges_event_id: 4, image: nil},
```
alkeshrails, line 2: This `id` issue may be occur in future and than we again need to fix it.
We should use 
`badges_event_ids = BadgesEvent.all.pluck(:id)`
and here we can use
`badges_event_ids[3]`

db/seeds/badges.rb
```diff
@@ -1,62 +1,62 @@
...
-  {name: "Admired", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Thank You", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Gives Back", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Empathetic", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Out of Love", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Higher Love", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Crazy in Love", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "First Mention", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
-  {name: "Welcome", description: nil, badge_type_id: nil, code: nil, badges_event_id: 11, image: nil},
+  {name: "New User of the Month", description: nil, badge_type_id: nil, code: nil, badges_event_id: 5, image: nil},
```
alkeshrails, line 8: `badges_event_ids[4]`
same as above.

db/seeds/badges.rb
```diff
@@ -1,62 +1,62 @@
...
+  {name: "gold-challenge-end", description: "Gold Medal winner for a challenge", badge_type_id: 1, code: "if max_rank <= 250\r\n  if participant_ranks>=0.2 && participant_ranks < 0.4\r\n    Participant.find_by(id: participant_id).add_badge('gold-challenge-end', {challenge_round: round_id})\r\n  end\r\nelsif max_rank > 250\r\n  if participant_ranks >=0.05 && participant_ranks < 0.1\r\n    Participant.find_by(id: participant_id).add_badge('gold-challenge-end', {challenge_round: round_id})\r\n  end\r\nend", badges_event_id: 6, image: nil},
+  {name: "bronze-challenge-end", description: "Bronze Medal winner for challenge", badge_type_id: 3, code: "if max_rank <= 250\r\n  if participant_ranks < 0.1\r\n    Participant.find_by(id: participant_id).add_badge('bronze-challenge-end', {challenge_round: round_id})\r\n  end\r\nelsif max_rank > 250\r\n  if participant_ranks < 0.01\r\n    Participant.find_by(id: participant_id).add_badge('bronze-challenge-end', {challenge_round: round_id})\r\n  end\r\nend\r\n", badges_event_id: 6, image: nil}
```
alkeshrails, line 61: `badges_event_ids[5]`
Same as above

db/seeds/badges.rb
```diff
@@ -1,62 +1,62 @@
 AicrowdBadge.create!([
```
piotrekpasciak, line 1: Seeds should use `find_or_create_by!()` method instead of `create!` so you can run them multiple times.

db/seeds/badges.rb
piotrekpasciak: Assign it like that to avoid this problem again:

```
third_badge =  BadgeType.find_by!(name: 'third')
fourth_event = BadgeEvent.find_by!(name: 'fourth')

{ badge_type: third_badge, badges_event: fourth_event }
```

PR #1485, opened 2020-05-19
adding category field for badges

nikhilrayaprolu:


app/models/aicrowd_badge.rb
```diff
@@ -1,5 +1,6 @@
 class AicrowdBadge < ApplicationRecord
   has_many :aicrowd_user_badges
+  belongs_to :badge_type
```
alkeshrails, line 3: Please add association of `AicrowdBadge` on model `BadgeType` as well. 
has_one or has_many, whatever is requirement.

PR #1487, opened 2020-05-20
Fix organisation new member form

piotrekpasciak:
Rollbar:
https://rollbar.com/aicrowd-production/aicrowd-production/items/369/

### DIFF
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -13,9 +13,15 @@ class MembersController < ApplicationController
   end
 
   def create
-    participant   = Participant.where(email: strong_params[:email]).first
-    flash[:error] = "No crowdAI participant can be found with that email address" if participant.blank?
-    flash[:info]  = "Participant added as an Organizer" if participant.present? && participant.organizers << @organizer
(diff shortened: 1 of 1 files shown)


app/controllers/members_controller.rb
```diff
@@ -13,9 +13,15 @@ def new
   end
 
   def create
-    participant   = Participant.where(email: strong_params[:email]).first
-    flash[:error] = "No crowdAI participant can be found with that email address" if participant.blank?
-    flash[:info]  = "Participant added as an Organizer" if participant.present? && participant.organizers << @organizer
+    participant           = Participant.where(email: strong_params[:email]).first
```
skbly7, line 16:
```
flash[:error] = "No crowdAI participant can be found with that email address" if participant.blank?
```

this error may not work after this refactor?

piotrekpasciak: We will have  "Participant does not exist" error message now.  Is it very important that error message would not change?

skbly7: Hmm ok, I see. In case this error message is coming it looks good. 😄

Issue #1473, opened 2020-05-15
Feature Request: Freezing Leaderboards


akshatcx:
Add an option to freeze a leaderboard x days before the challenge ends.
- New submissions entertained but leaderboard won't be updated
- Ability to see only your own submissions (others are hidden) after the freeze
- Organizers can see the updated leaderboard even if frozen

skbly7:
Will be implemented by #1493

PR #1493, opened 2020-05-20
Freeze leaderboard Feature

alkeshrails:
1. Freeze leaderboard enable, disable radio with duration field and migration

### DIFF
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -168,7 +168,7 @@ class LeaderboardsController < ApplicationController
   end
 
   def freeze_leaderbaord_participant?
-    freeze_condition? && get_my_leaderboard.pluck(:submitter_id).include?(current_participant.id)
+    current_participant.present? && freeze_condition? && get_my_leaderboard.pluck(:submitter_id).include?(current_participant.id)
   end
 end
 
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -140,14 +140,9 @@ class LeaderboardsController < ApplicationController
   end
 
   def freeze_condition?
-    return false if current_participant&.admin? || participant_is_organizer
+    return false if current_participant&.admin? || policy(@challenge).edit?
 
     @current_round.freeze_flag && freeze_time(@current_round)
-  end
-
-  def participant_is_organizer
-    organizers_email = @challenge.organizers.flat_map { |organizer| organizer.participants }.pluck(:email)
-    organizers_email.include?(current_participant&.email)
   end
 
   def freeze_time(ch_round)
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -34,6 +34,10 @@ class LeaderboardsController < ApplicationController
     @post_challenge   = post_challenge?
     @following        = following?
 
+    if freeze_leaderbaord_participant?
+      @my_leaderboard = get_my_leaderboard
+    end
+
     unless is_disentanglement_leaderboard?(@leaderboards.first)
       @countries = @filter.call('participant_countries')
       @affiliations = @filter.call('participant_affiliations')
@@ -150,6 +154,21 @@ class LeaderboardsController < ApplicationController
     return false if ch_round.end_dttm.nil? || (ch_round.end_dttm - Time.now.utc).negative?
 
     ch_round.end_dttm - Time.now.utc < ch_round.freeze_duration * 60 * 60
+  end
+
+  def get_my_leaderboard
+    filter = { challenge_round_id: @current_round&.id.to_i, meta_challenge_id: nil, freeze_leaderboard: false, submitter_type: "Participant", submitter_id: current_participant.id }
+    @get_my_leaderboard ||=  if post_challenge?
+                              policy_scope(OngoingLeaderboard)
+                                .where(filter)
+                            else
+                              policy_scope(Leaderboard)
+                                .where(filter)
+                            end
+  end
+
+  def freeze_leaderbaord_participant?
+    freeze_condition? && get_my_leaderboard.pluck(:submitter_id).include?(current_participant.id)
   end
 end
 
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -103,7 +103,7 @@ class LeaderboardsController < ApplicationController
       DisentanglementLeaderboard
         .where(challenge_round_id: @current_round)
         .freeze_record(current_participant)
-    elsif post_challenge? || policy(@challenge).edit? || current_participant.admin
+    elsif post_challenge? || policy(@challenge).edit? || current_participant&.admin
       policy_scope(OngoingLeaderboard)
         .where(filter)
     else
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -33,6 +33,7 @@ class LeaderboardsController < ApplicationController
     @challenge_rounds = @challenge.challenge_rounds.started
     @post_challenge   = post_challenge?
     @following        = following?
+
     unless is_disentanglement_leaderboard?(@leaderboards.first)
       @countries = @filter.call('participant_countries')
       @affiliations = @filter.call('participant_affiliations')
@@ -93,13 +94,17 @@ class LeaderboardsController < ApplicationController
   end
 
   def set_leaderboards
-    filter = {challenge_round_id: @current_round&.id.to_i, meta_challenge_id: nil}
+    filter = { challenge_round_id: @current_round&.id.to_i, meta_challenge_id: nil }
+
+    filter[:freeze_leaderboard] = freeze_condition?
+
     if @meta_challenge.present?
       filter[:meta_challenge_id] = @meta_challenge.id
     end
     @leaderboards = if @challenge.challenge == "NeurIPS 2019 : Disentanglement Challenge"
       DisentanglementLeaderboard
         .where(challenge_round_id: @current_round)
+        .freeze_record(current_participant)
     elsif post_challenge?
       policy_scope(OngoingLeaderboard)
         .where(filter)
@@ -129,4 +134,22 @@ class LeaderboardsController < ApplicationController
   def is_disentanglement_leaderboard?(leaderboard)
     leaderboard.class.name == 'DisentanglementLeaderboard'
   end
+
+  def freeze_condition?
+    return false if current_participant&.admin? || participant_is_organizer
+
+    @current_round.freeze_flag && freeze_time(@current_round)
+  end
+
+  def participant_is_organizer
+    organizers_email = @challenge.organizers.flat_map { |organizer| organizer.participants }.pluck(:email)
+    organizers_email.include?(current_participant&.email)
+  end
+
+  def freeze_time(ch_round)
+    return false if ch_round.end_dttm.nil? || (ch_round.end_dttm - Time.now.utc).negative?
+
+    ch_round.end_dttm - Time.now.utc < ch_round.freeze_duration * 60 * 60
+  end
 end
+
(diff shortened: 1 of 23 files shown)


app/models/challenge_round.rb
```diff
@@ -32,6 +32,7 @@ class ChallengeRound < ApplicationRecord
   validates :submission_limit_period, presence: true
   validates :primary_sort_order, presence: true
   validates :secondary_sort_order, presence: true
+  validate :positive_integer
```
piotrekpasciak, line 35: It's better to use Rails build in validations e.g:
- `validates :freeze_duration, numericality: { greater_than: 0 }, if: :present?`

app/models/concerns/freeze_record.rb
```diff
@@ -0,0 +1,31 @@
+require 'active_support/concern'
```
piotrekpasciak, line 1: Are you sure this `require` is needed?

app/models/concerns/freeze_record.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module FreezeRecord
+  extend ActiveSupport::Concern
+
+  included do
+    def self.freeze_record(participant)
+      return all if !participant.present? || participant_is_organizer(participant.email) || all.where(submitter_type: "Participant").pluck(:submitter_id).exclude?(participant.id)
```
piotrekpasciak, line 8: Let's avoid `!` when possible as it's hard to read them:
- `participant.blank?`

app/models/submission.rb
```diff
@@ -116,6 +116,19 @@ def as_json(options = {})
     )
   end
 
+  def self.freeze_record(participant, current_round, challenge)
+    return all if !participant.present? || participant_is_organizer(participant.email, challenge) || all.pluck(:participant_id).exclude?(participant.id)
```
piotrekpasciak, line 120: `participant.blank?`

piotrekpasciak:
@alkeshrails @skbly7 This looks quite complicated. What's business logic behind it? 

Why can't we?
- click freeze leaderboard and set freeze time amount
- stop calculating new leaderboards when challenge_round / challenge is frozen

We could simply stop calculating frozen leaderboards with one line here:
- https://github.com/AIcrowd/AIcrowd/blob/master/app/jobs/calculate_leaderboard_job.rb#L4

ThierryDeruyttere:
Is there an ETA when this feature will be finished? We would like to use it for the last two weeks of our challenge.
Also, it would be nice if there could be an option that the leaderboard remains frozen even after the challenge is ended. 
The reason for that is that we are interested to show the final ranking at our workshop. After the workshop we would unfreeze the leaderboard.

app/services/challenge_rounds/create_leaderboards_service.rb
```diff
@@ -241,5 +245,17 @@ def assign_row_num(leaderboards)
     def window_border_dttm(post_challenge)
       (submissions.find_by(post_challenge: post_challenge, meta_challenge_id: meta_challenge_id)&.created_at || Time.current) - challenge_round.ranking_window.hours
     end
+
+    def create_freeze_leaderabord(challenge_round, leaderboard_type)
```
piotrekpasciak, line 249: This significantly differs from:
https://github.com/AIcrowd/AIcrowd/blob/master/app/services/challenge_rounds/create_leaderboards_service.rb#L90

Is it part of requirements to drop `previous_rank` calculation?

app/services/challenge_rounds/create_leaderboards_service.rb
```diff
@@ -25,11 +25,13 @@ def call
         previous_leaderboards = build_base_leaderboards('leaderboard', [false], window_border_dttm([false]))
 
         create_leaderboards(leaderboards, previous_leaderboards)
+        create_freeze_leaderabord(challenge_round, 'leaderboard')
```
piotrekpasciak, line 28: During the standup we talked about modifying `build_base_leaderboards`.
https://github.com/AIcrowd/AIcrowd/blob/master/app/services/challenge_rounds/create_leaderboards_service.rb#L52 

Is there any particular reason why you chose to define `create_freeze_leaderabord` method?

This class was optimised for performance and it would be good idea to keep it optimized.

app/services/challenge_rounds/create_leaderboards_service.rb
piotrekpasciak, line 249: Typo: `leaderabord` should be `leaderboard`

app/services/challenge_rounds/create_leaderboards_service.rb
```diff
@@ -25,11 +25,13 @@ def call
         previous_leaderboards = build_base_leaderboards('leaderboard', [false], window_border_dttm([false]))
```
piotrekpasciak, line 25: Please write test coverage for new functionality of this class here:
https://github.com/AIcrowd/AIcrowd/blob/master/spec/services/challenge_rounds/create_leaderboards_service_spec.rb

skbly7:
Hi @ThierryDeruyttere, this feature is on the last round of code review and should be available this week (hopefully by tomorrow end).

Yes, we have taken care of it, working would be as follows:
1. Organizers can enable freeze along with X-hours they want to freeze (from the round end)
2. Participants wouldn't be able to see scores for co-participant submissions as well (made in freeze time) and self-scores will be visible, so they can't deduce leaderboard.
3. The freeze will remain active until organizers manually toggle/disable "freeze"
4. The leaderboard will remain "non-frozen" for organizers, so they can view the live leaderboard and plan things accordingly before undo-freezing/results-announcement.

Do let us know here or in #1473 for any scenario we may have missed and you would like to see it. 😃

app/services/challenge_rounds/create_leaderboards_service.rb
piotrekpasciak: Akhilesh said that: `create a separate method that create Freeze Leaderboard for me.
So I show Freeze leaderboard to participant and latest leaderboard to organiser.`

app/controllers/leaderboards_controller.rb
```diff
@@ -129,4 +138,37 @@ def submitter_submissions(submitter)
...
+  end
+
+  def freeze_time(ch_round)
+    return false if ch_round.end_dttm.nil? || (ch_round.end_dttm - Time.now.utc).negative?
+
+    ch_round.end_dttm - Time.now.utc < ch_round.freeze_duration * 60 * 60
+  end
+
+  def get_my_leaderboard
+    filter = { challenge_round_id: @current_round&.id.to_i, meta_challenge_id: nil, freeze_leaderboard: false, submitter_type: "Participant", submitter_id: current_participant.id }
```
skbly7, line 160: Why are we doing this filter here instead of something in `set_leaderboard`? Looks like duplication of too many params and not proper approach.

skbly7: Basically check for `policy(...).edit?` and accordingly override policy_scope

app/views/leaderboards/_leaderboards.html.erb
```diff
@@ -3,13 +3,14 @@
     participants = leaderboard_participants(leaderboard)
     challenge    = leaderboard.challenge
     truncate     = truncate.present? ? truncate : false
+    hidden_rank  = hidden_rank||=nil
```
skbly7, line 6: We shouldn't use too many different variables for technically same concept, you can use something like value of `freeze_condition?` here?

app/views/leaderboards/index.html.erb
```diff
@@ -89,6 +89,7 @@
               </tr>
               </thead>
               <tbody id="leaderboards-div" data-page='1' data-request-completed='true' data-controller="load-more" data-action="scroll@window->load-more#load_more_data" data-meta-challenge-id="<%= @challenge.is_a_problem? ? @challenge.challenge_problem.challenge.friendly_id : '' %>">
+                <%= render partial: 'leaderboards', locals: { leaderboards: @my_leaderboard, current_round: @current_round, truncate: 'true', hidden_rank: true } if @my_leaderboard.present? %>
```
skbly7, line 92: Why extra partial?

app/services/challenge_rounds/create_leaderboards_service.rb
```diff
@@ -241,5 +246,17 @@ def assign_row_num(leaderboards)
     def window_border_dttm(post_challenge)
       (submissions.find_by(post_challenge: post_challenge, meta_challenge_id: meta_challenge_id)&.created_at || Time.current) - challenge_round.ranking_window.hours
     end
+
```
skbly7, line 249: Is `create_freeze_leaderboard` required at all and all other logic changes in this file?

Why can't we just use `hallenge_round.freeze_flag && challenge_round.freeze_duration.positive?` in `window_border_dttm`?

Other things kind of remain as it is, given we are using `OngoingLeaderboard` and `Leaderboard` for display purposes in `leaderboard_controller`.

app/controllers/leaderboards_controller.rb
alkeshrails: Actually this method is diff to set_leaderboard.
set_leaderboard get the FreezeLeaderboard record and show.
this method get current score from leaderboard. and we show this in first row to participant without his rank.
Participant that submission is in Freeze duration, will get 2 diff leaderbaord.
First is Freeze Leaderboard to set_leaderboard.
Second his current score with updated leaderboard to get_my_leaderboard.

app/services/challenge_rounds/create_leaderboards_service.rb
alkeshrails: We need 2 leaderbaord.
One is updated leaderboard to show Admin and Organiser.
Second is Freeze Leaderbaord to show participant.
So I create freezze_leaderbaord separately.

app/views/leaderboards/_leaderboards.html.erb
alkeshrails: for participant freeze_condition is true for all (his own record and for other record).
I need to hide just his own rank, so I set `hidden_rank` at `my_leaderboard` called this partial.

app/views/leaderboards/index.html.erb
alkeshrails: we have 2 diff leaderbaord.
One is his own leaderoard that come from latest leaderboard to show his own score (without rank).
Second is FreezeLeaderboard.

app/controllers/leaderboards_controller.rb
```diff
@@ -119,4 +128,37 @@ def submitter_submissions(submitter)
...
   end
+
+  def freeze_condition?
+    return false if current_participant&.admin? || participant_is_organizer
+
+    @current_round.freeze_flag && freeze_time(@current_round)
+  end
+
+  def participant_is_organizer
+    organizers_email = @challenge.organizers.flat_map { |organizer| organizer.participants }.pluck(:email)
```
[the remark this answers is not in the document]
alkeshrails: oh yes, Thanks.
Actually first I used concern so it was there.

ThierryDeruyttere:
I have a question about the freezing of the leaderboard feature.
As of now, if one wants to freeze the leaderboard, you need to specify a value in the freeze window textbox.
This value represents the number of hours before the finish of the challenge that the leaderboard should be frozen (see image).
In my opinion, this was not clear at all and felt very counter-intuitive.
Why can't we just put in the date and time when we want the leaderboard to be frozen?
This would be easier for anyone to use this feature. 

image

ThierryDeruyttere:
@alkeshrails

ThierryDeruyttere:
Also, when exporting the leaderboard to json in the organisers menu will still give the results of the frozen leaderboard. This might not be what you'd want. We would like to get the final leaderboard without unfreezing it.

skbly7:
@ThierryDeruyttere yes, correct. We want the organizers to be able to download the submissions/leaderboard, we will fix the export.

skbly7:
> This value represents the number of hours before the finish of the challenge that the leaderboard should be frozen (see image).

Thanks for the suggestion.

A lot of times challenge end time moves up or down based on scenarios and we felt number of hours may be a better way for this feature. I will recheck with team on it and we will update it accordingly.

PR #1496, opened 2020-05-21
Drop dependency on Mandrill

piotrekpasciak:
Monday:
https://aicrowd.monday.com/boards/363093650/pulses/569692370


app/controllers/api/external_graders_controller.rb
```diff
@@ -394,7 +394,9 @@ def participant_qualified?(challenge, participant)
   end
 
   def notify_admins(submission)
-    Admin::SubmissionNotificationJob.perform_later(submission)
+    Participant.admins.with_every_email_preference.each do |participant|
```
alkeshrails, line 397: We should do this by a Job.

app/controllers/challenge_call_responses_controller.rb
```diff
@@ -11,10 +11,13 @@ def create
...
-          .perform_later(@challenge_call_response)
-      redirect_to challenge_call_show_path(
-        @challenge_call, @challenge_call_response)
+      participants = @challenge_call_response.challenge_call&.organizer&.participants
+
+      participants&.each do |participant|
+        Admin::NotificationsMailer.challenge_call_response_email(participant, @challenge_call_response).deliver_later
+      end
+
+      redirect_to challenge_call_show_path(@challenge_call, @challenge_call_response)
```
alkeshrails, line 20: Redirect with message.

app/controllers/organizer_applications_controller.rb
```diff
@@ -3,8 +3,12 @@ class OrganizerApplicationsController < ApplicationController
 
   def create
     organizer_application = OrganizerApplication.create!(organizer_application_params)
-    Admin::OrganizerApplicationNotificationJob.perform_later(organizer_application)
-    OrganizerApplicationNotificationJob.perform_later(organizer_application)
+
+    Participant.admins.each do |participant|
+      Admin::NotificationsMailer.organizer_application_notification_email(participant, organizer_application).deliver_later
```
alkeshrails, line 8: Should manage this by a Job.

app/controllers/participant_clef_tasks_controller.rb
```diff
@@ -49,7 +49,10 @@ def set_status(clef_task, participant_clef_task)
     if clef_task.eua_required?
       if participant_clef_task.eua_file.present?
         participant_clef_task.update(status_cd: 'submitted')
-        Organizer::EuaNotificationJob.perform_later(clef_task.id, current_participant.id)
+
+        clef_task.organizer.participants.where(clef_email: true).each do |organizer_participant|
```
alkeshrails, line 53: Should manage by a Job.

app/controllers/submissions_controller.rb
```diff
@@ -305,7 +305,9 @@ def set_submissions_remaining
   end
 
   def notify_admins
-    Admin::SubmissionNotificationJob.perform_later(@submission)
+    Participant.admins.with_every_email_preference.each do |participant|
+      Admin::NotificationsMailer.submission_notification_email(participant, @submission).deliver_later
```
alkeshrails, line 309: I think we should manage all this type of mail deliver to Job. We are running each loop so may be its increase response time.
And we already use deliver_later so Job is good for this type loop mailer.
Please check on other place also.

app/helpers/invitations_helper.rb
```diff
@@ -0,0 +1,61 @@
+module InvitationsHelper
+  def invitation_amount_of_other_people(team)
+    other_team_member_count = team.team_participants.count - 1
```
alkeshrails, line 3: Just a quick question. Value of `team.team_participants.count` never 0 or 1 in any case?

app/mailers/participants/invitations_mailer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      mail(to: @participant.email, subject: subject)
+    end
+
+    def invitation_canceled_email(team_invitation)
+      set_participant_from_invitee(team_invitation.invitee)
+      @team = team_invitation.team
+      @email_preferences_url = EmailPreferencesTokenService.new(@participant).preferences_token_url
+      @notification_reason   = 'Someone canceled your invitation to a team.'
+      subject                = '[AIcrowd] Your Invitation Was Canceled'
```
alkeshrails, line 20: We should use team name as well in subject and notification_reason.
Same for other places.

app/helpers/invitations_helper.rb
piotrekpasciak: Not sure, but it doesn't make sense to change it here, as e-mail text expects that there are other people in team.

app/mailers/participants/invitations_mailer.rb
piotrekpasciak: i've updated other subjects.

PR #1504, opened 2020-05-25
Feature/notification

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/575007944


app/services/notification_service.rb
```diff
@@ -0,0 +1,112 @@
...
+    @participant       = Participant.find(participant_id)
+    @notifiable        = notifiable
+    @notification_type = notification_type
+  end
+
+  def call
+    eval(@notification_type)
+  end
+
+  def comment
```
piotrekpasciak, line 15: Shouldn't those metthods be private?

app/services/notification_service.rb
```diff
@@ -0,0 +1,112 @@
...
+  include ActionView::Helpers::AssetTagHelper
+
+  def initialize(participant_id, notifiable, notification_type)
+    @participant       = Participant.find(participant_id)
+    @notifiable        = notifiable
+    @notification_type = notification_type
+  end
+
+  def call
+    eval(@notification_type)
```
piotrekpasciak, line 12: Can we avoid using `eval`? It's very dangerous method that. can open application. for Code injections  attacks.

app/controllers/application_controller.rb
```diff
@@ -26,6 +27,12 @@ def append_info_to_payload(payload)
     payload[:user_id]    = current_user.id if current_user
   end
 
+  def notifications
+    if current_user.present?
+      @notifications = current_user.notifications
```
piotrekpasciak, line 32: `current_user&.notifications` - would be shorter version of this method

app/controllers/participants_controller.rb
```diff
@@ -97,6 +97,23 @@ def remove_image
                 notice: 'Image removed.'
   end
 
+  def discourse_notifications
```
piotrekpasciak, line 100: json endpoints should be under `api/v1/` namespace, because of convention and performance.

app/controllers/participants_controller.rb
```diff
@@ -97,6 +97,23 @@ def remove_image
...
+    Discourse::NotificationService.new(notification: params['notification']).call
+    render json: { message: 'creating' }, status: :ok
+  end
+
+  def notifications_message
```
piotrekpasciak, line 105: You've already. defined. this method in `application_contrroller.rb` and this class. inherits. from it. You should be able to remove it from here.

app/controllers/participants_controller.rb
```diff
@@ -97,6 +97,23 @@ def remove_image
...
+    @notifications = current_user.notifications
+  end
+
+  def read_notification
+    @notification = current_user.notifications.find(params[:id])
+    @notification.update(is_new: false)
+
+    redirect_url = @notification.notification_url.present? ? @notification.notification_url : root_url
+    redirect_to redirect_url
```
piotrekpasciak, line 114: `redirect_to @notification.notification_url.presence || root_url`

app/helpers/application_helper.rb
```diff
@@ -147,10 +147,48 @@ def footer_text
...
+
   def social_image_path(site)
     "/assets/img/icon-#{site}.svg"
   end
 
+  def check_notification_status(notifications)
+    notifications&.pluck(:is_new)&.uniq&.any?(true) ? 'notification-bell-icon-active' : 'notification-bell-icon'
+  end
+
+  def is_new?(notification)
```
piotrekpasciak, line 188: `is_new?` is ambiguous name for this method. please rename it to something like `notification_new_class`.

app/services/discourse/notification_service.rb
```diff
@@ -0,0 +1,71 @@
+module Discourse
+  class NotificationService < ::Discourse::BaseService
```
piotrekpasciak, line 2: We need test coverage for this.  class.

app/services/notification_service.rb
```diff
@@ -0,0 +1,69 @@
+class NotificationService
```
piotrekpasciak, line 1: We need test coverage for this class.

app/services/notification_service.rb
```diff
@@ -0,0 +1,69 @@
...
+  include ActionView::Helpers::AssetTagHelper
+
+  def initialize(participant_id, notifiable, notification_type)
+    @participant       = Participant.find(participant_id)
+    @notifiable        = notifiable
+    @notification_type = notification_type
+  end
+
+  def call
+    send(@notification_type) if ['graded', 'failed', 'leaderboard'].include?(@notification_type) && @participant.present?
```
piotrekpasciak, line 12: `public_send` will be sufficient here.

piotrekpasciak, line 12: You can remove. `&& @participant.present?` as ` Participant.find(participant_id)`. will. throw exception if it's. not.

app/controllers/application_controller.rb
alkeshrails: Done

app/controllers/participants_controller.rb
alkeshrails: Done

app/controllers/participants_controller.rb
alkeshrails: Yes agree, but I don't have route and view page for application controller's `notification` method.
and this method has route and view to show all Notifications.
Application controller method is only load notification to show on dropdown.

app/controllers/participants_controller.rb
alkeshrails: Done

app/helpers/application_helper.rb
piotrekpasciak: `unread_notification?`  would be good

app/controllers/participants_controller.rb
piotrekpasciak: Wasn't aware this is used. Looks good. then.

app/services/notification_service.rb
alkeshrails: Done

app/controllers/participants_controller.rb
alkeshrails: Thanks

app/helpers/application_helper.rb
alkeshrails: Okay

app/services/notification_service.rb
alkeshrails: make it public_send, now I need to remove `private`

piotrekpasciak: Keep. `send`. I'll test it later.

alkeshrails: Okay

PR #1507, opened 2020-05-28
Fixing Badges Icon, Removing the Badges Tab and Fixing the Scroll on Profile page, Variation in Ratings

nikhilrayaprolu:


app/views/participants/_achievements.erb
```diff
@@ -32,9 +32,40 @@
...
-        </div>
-      </div>
+      <% end %>
+    </div>
   </div>
-  </div>
-</section>
\ No newline at end of file
+</section>
+<div class="container-fluid" style="<%= style ||= '' %>">
```
piotrekpasciak, line 39: Why do you use `<%= style ||= '' %>` ?

`<%= style >` should work as well.

PR #1511, opened 2020-06-02
Fix challenge status filter issue

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/589241471
The issue was in services/terminate_challenges.rb because in existing code featured_sequence is getting updated with value 0 in same file and there is a validation for featured_sequence in challenge model for 0 value.


alkeshrails:
@skbly7 As we're validating featured sequence for 0 value in challenge model.

PR #1516, opened 2020-06-03
adding new fixes for rating

nikhilrayaprolu:


app/assets/awards/award-other.svg
```diff
@@ -0,0 +1,16 @@
+<svg viewBox="0 0 34 40" xmlns="http://www.w3.org/2000/svg">
```
skbly7, line 1: Remove extra file

db/migrate/20200604112701_add_default_to_badge_type_in_badges.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultToBadgeTypeInBadges < ActiveRecord::Migration[5.2]
```
skbly7, line 1: Why does this migration exist? It isn't being used anywhere.

PR #1542, opened 2020-06-08
Minor CSS change for card-footer (organisers display)

skbly7:
Before:
image

After:
image


app/assets/stylesheets/application-redesign.css.scss
```diff
@@ -4399,7 +4399,6 @@ input[type="button"].btn-block {
 
 .card-challenge .card-footer a {
   color: #292C2D;
-  display: flex;
   align-items: center;
```
piotrekpasciak, line 4402: You should also remove. `align-items: center;` as this property won't work without `flexbox`

skbly7: removed now

PR #1544, opened 2020-06-08
Add code block for admin ckeditor

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/591919188


app/admin/blog.rb
```diff
@@ -30,7 +30,7 @@ def permitted_params
       f.input :seq
       f.input :published
       f.label 'Body', class: 'ckeditor_label blog'
-      f.text_area :body, class: 'ckeditor'
+      f.cktext_area :body
```
skbly7, line 33: Will the data be deleted?

alkeshrails: No, it will show data to DB in editor. but this implementation has issue with list attribute, so I'm checking that.

skbly7: My bad, thanks

PR #1549, opened 2020-06-09
Few fixes to the rating system and drop down fix for the profile page

nikhilrayaprolu:


app/assets/images/awards/award-gold.svg
```diff
@@ -1,11 +1,16 @@
-<svg viewBox="0 0 30 40" xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="0 0 34 40" xmlns="http://www.w3.org/2000/svg">
```
skbly7, line 1: This image is being used in other places not related to medals too!

Please add a new image (different name) instead.

PR #1559, opened 2020-06-12
Add hide submissions option

nickinack:
----- WORK IN PROGRESS ----

Problems:
1. LeaderBoard Service does not update.

Working:
1. Effectively hides submissions from submissionms page
2. Hides Submissions in the challenges landing page.


app/views/submissions/_submissions.html.erb
```diff
@@ -3,37 +3,39 @@
     participant     = submission.participant
     challenge       = submission.challenge
     challenge_round = submission.challenge_round
+    visible         = submission.visible
```
skbly7, line 6: It isn't required if you are doing `visible: true` in the `submissions_controller`

app/views/challenges/show.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= render partial: 'shared/challenges/masthead', locals: { challenge: @challenge, challenge_rounds: @challenge_rounds, vote: @vote, follow: @follow } %>
 <%= render partial: 'challenges/show/subnav', locals: { challenge: @challenge } %>
 
+
```
skbly7, line 4: Remove the files with no changes from your PR

PR #1561, opened 2020-06-15
Add UUID suffix to Eua files

piotrekpasciak:


app/uploaders/eua_uploader.rb
```diff
@@ -5,6 +5,10 @@ def store_dir
     "EUAs/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
   end
 
+  def filename
+    "#{original_filename.to_s.split('.').first}-#{SecureRandom.uuid}.#{file&.extension}"
+  end
```
alkeshrails, line 10: where its call and get `original_filename` and `file` value?

skbly7: @alkeshrails  It is coming from CarrierWave class. https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Customize-your-version-file-names

@piotrekpasciak might be better to use `File.basename` or something here?

piotrekpasciak: I'll check how `File.basename` works.

piotrekpasciak: Where did you get `File.basename`? There is nothing like that.

skbly7:
```
>> file = "/home/gumby/work/ruby_file.with.dot.rb" #=> "/home/gumby/work/ruby_file.with.dot.rb"
>> File.basename(file, File.extname(file)) #=> "ruby_file.with.dot"
```

Reference: https://stackoverflow.com/a/40846364/2529149

PR #1562, opened 2020-06-15
Fix params in CalculateLeaderboardJob

piotrekpasciak:


app/jobs/calculate_leaderboard_job.rb
```diff
@@ -7,10 +7,10 @@ def perform(challenge_round_id:)
     elsif ChallengeRound.find(challenge_round_id).challenge.meta_challenge
       CalculateMetaLeaderboardService.new(challenge_id: ChallengeRound.find(challenge_round_id).challenge.id).call
     else
-      ChallengeRounds::CreateLeaderboardsService.new(challenge_round_id: challenge_round_id).call
```
skbly7, line 10: Hey, it will be easier to keep it as `challenge_round_id` to keep consistency with other `Service.new`?

piotrekpasciak: Nearly every Service class takes object as parameter thus current version is more consistent then your proposition.

Only jobs are required to take id and reload the object.

skbly7: cool works until the decision have been thought upon!

PR #1566, opened 2020-06-16
Hide-Submission-Features-added

nickinack:
Hide submission feature along with leaderboard changes has been added.

### DIFF
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -15,7 +15,7 @@ class LeaderboardsController < ApplicationController
         @submitter_submissions.merge!("submitter#{submitter.id}_submissions_by_day": submitter_submissions(submitter).group_by_created_at) if submitter.present?
       end
     end
-    @top_three_winners = @leaderboards.where(baseline: nil).first(3)
+    @top_three_winners = @leaderboards.where(baseline: false).first(3)
     if params[:country_name].present? || params[:affiliation].present?
       @leaderboards = @leaderboards.where(id: @filter.call('leaderboard_ids'))
       @leaderboards = paginate_leaderboards_by(:seq)
--- a/app/services/challenge_rounds/create_leaderboards_service.rb
+++ b/app/services/challenge_rounds/create_leaderboards_service.rb
@@ -53,25 +53,29 @@ module ChallengeRounds
       users_leaderboards = []
 
       teams_submissions(post_challenge, cuttoff_dttm).each do |team_id, submissions|
-        first_graded_submission = submissions.find { |submission| submission.grading_status_cd == 'graded' }
+        first_graded_submission = submissions.find { |submission| (submission.grading_status_cd == 'graded' && submission.visible)  }
         next if first_graded_submission.blank?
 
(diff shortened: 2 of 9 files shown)


app/services/challenge_rounds/create_leaderboards_service.rb
```diff
@@ -53,25 +53,29 @@ def build_base_leaderboards(leaderboard_type, post_challenge, cuttoff_dttm)
       users_leaderboards = []
```
piotrekpasciak, line 53: Instead of all changes in this file you should add a single change here:
https://github.com/AIcrowd/AIcrowd/blob/master/app/services/challenge_rounds/create_leaderboards_service.rb#L6

`where(visible: true)`

piotrekpasciak, line 53: Please add new test scenario for hidden submissions here:
- https://github.com/AIcrowd/AIcrowd/blob/master/spec/services/challenge_rounds/create_leaderboards_service_spec.rb#L3

app/views/submissions/_submissions.html.erb
```diff
@@ -3,37 +3,39 @@
     participant     = submission.participant
     challenge       = submission.challenge
     challenge_round = submission.challenge_round
+    visible         = submission.visible
```
piotrekpasciak, line 6: Remove all `visible` changes from this file. You should list only visible submissions instead. This should be done in controller:
https://github.com/AIcrowd/AIcrowd/blob/master/app/controllers/submissions_controller.rb#L63
https://github.com/AIcrowd/AIcrowd/blob/master/app/controllers/submissions_controller.rb#L70

db/migrate/20200613094254_add_visible_to_submissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisibleToSubmissions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :submissions, :visible, :boolean, default: true
```
piotrekpasciak, line 3: It's good practice to. add. `null: false` constraint to boolean columns. to reduce amount of bugs.  Please add it.

db/migrate/20200613094254_add_visible_to_submissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisibleToSubmissions < ActiveRecord::Migration[5.2]
+  def change
```
piotrekpasciak, line 2: Please add `schema.rb` with changes from this migration. Otherwise. we will have trouble to deploy your branch.

app/controllers/leaderboards_controller.rb
```diff
@@ -15,7 +15,7 @@ def index
         @submitter_submissions.merge!("submitter#{submitter.id}_submissions_by_day": submitter_submissions(submitter).group_by_created_at) if submitter.present?
       end
     end
-    @top_three_winners = @leaderboards.where(baseline: nil).first(3)
+    @top_three_winners = @leaderboards.where(baseline: false).first(3)
```
skbly7, line 18: It needs to be `nil` afaik and not `false`

PR #1571, opened 2020-06-16
Notification feature : Feedback changes

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/610004882

### DIFF
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -103,13 +103,11 @@ class ParticipantsController < ApplicationController
 
   def read_notification
     if params[:ids].present?
-      @notifications = current_user.notifications.find(JSON.parse(params[:ids]))
-      @notifications.each do |notification|
-        notification.update(is_new: false)
-      end
+      @notifications = current_user.notifications.where(id: JSON.parse(params[:ids]))
+      @notifications.update_all(is_new: false)
     else
       @notification = current_user.notifications.find(params[:id])
-      @notification.update(is_new: false)
+      @notification.update!(is_new: false)
     end
 
     redirect_to @notification.notification_url || root_url
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -102,8 +102,15 @@ class ParticipantsController < ApplicationController
   end
 
   def read_notification
-    @notification = current_user.notifications.find(params[:id])
-    @notification.update(is_new: false)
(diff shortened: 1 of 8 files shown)


app/services/notification_service.rb
```diff
@@ -44,19 +45,21 @@ def failed
         message:           message,
         thumbnail_url:     thumb,
         notification_url:  link,
+        challenge_id:      @notifiable.challenge.id,
         is_new:            true)
   end
 
   def leaderboard
     return if @participant.nil?
```
skbly7, line 53: return if @notifiable.previous_row_num == @notifiable.row_num

app/controllers/participants_controller.rb
```diff
@@ -102,8 +102,15 @@ def notifications_message
   end
 
   def read_notification
-    @notification = current_user.notifications.find(params[:id])
-    @notification.update(is_new: false)
+    if params[:ids].present?
+      @notifications = current_user.notifications.find(JSON.parse(params[:ids]))
+      @notifications.each do |notification|
```
piotrekpasciak, line 107: Use either `update!` to make sur everything works or `update_all` to make action faster. It doesn't make sense to validate objects and do nothing about validations.

app/controllers/participants_controller.rb
```diff
@@ -102,8 +102,15 @@ def notifications_message
...
+        notification.update(is_new: false)
+      end
+    else
+      @notification = current_user.notifications.find(params[:id])
+      @notification.update(is_new: false)
```
piotrekpasciak, line 112: `update!`

app/services/notification_service.rb
```diff
@@ -28,13 +32,18 @@ def graded
...
         is_new:            true)
   end
 
   def failed
     message = "Your #{@notifiable.challenge.challenge} Challenge submission ##{@notifiable.id} failed to evaluate."
     thumb   = @notifiable.challenge.image_file.url
-    link    = challenge_submissions_url(@notifiable.challenge, my_submissions: true)
+    link    = challenge_submission_url(@notifiable.challenge, @notifiable.id)
+
+    existing_notification = @participant.notifications.where(notification_type: "failed", notifiable_type: "Submission", message: message, is_new: true).first
```
skbly7, line 44: The message and type doesn't really need to be same for returning.

Can we check if *any* `notifiable_type: "Submission"` is present for this submission ID and return if true?
(same for graded function above)

PR #1573, opened 2020-06-16
Hide submissions option

nickinack:


app/controllers/leaderboards_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
         @submitter_submissions.merge!("submitter#{submitter.id}_submissions_by_day": submitter_submissions(submitter).group_by_created_at) if submitter.present?
       end
     end
-    @top_three_winners = @leaderboards.where(baseline: nil).first(3)
+    @top_three_winners = @leaderboards.where(baseline: false).first(3)
```
piotrekpasciak, line 20: You still need to talk with Shivam about this change. He said previously. that. you shouldn't. make that change.

PR #1584, opened 2020-06-19
Fix EUA files suffix issue

piotrekpasciak:


app/uploaders/eua_uploader.rb
```diff
@@ -6,9 +6,9 @@ def store_dir
   end
 
   def filename
-    full_name = original_filename.to_s
-
-    "#{File.basename(full_name, File.extname(full_name))}_#{SecureRandom.uuid}.#{file&.extension}"
+    if original_filename.present?
+      @unique_filename ||= "#{File.basename(original_filename.to_s, File.extname(original_filename.to_s))}_#{SecureRandom.uuid}.#{file&.extension}"
```
skbly7, line 10: What about different file name coming when uploaded vs links rendering/DB?

Is it because of SecureRandom thing, can be seed it instead with anything so it remains deterministic for same file?

piotrekpasciak: To handle that situation I'm using memoization:
- @unique_filename ||=

PR #1585, opened 2020-06-19
Prevent to create notification with existing notfication message

alkeshrails:
run rake task to delete notification of leaderboard that don't have `show_leaderboard?` permission
`rake remove_notification:leaderbaord_is_not_for_show`
This rake task also delete notification that's leaderboard is not exist in DB.

### DIFF
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -20,7 +20,7 @@ class NotificationService
     thumb   = @notifiable.challenge.image_file.url
     link    = challenge_submission_url(@notifiable.challenge, @notifiable.id)
 
-    existing_notification = @participant.notifications.where(notifiable: @notifiable, is_new: true).first
+    existing_notification = @participant.notifications.where(notifiable: @notifiable).first
 
     return if existing_notification.present?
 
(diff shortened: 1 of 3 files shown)


app/services/notification_service.rb
```diff
@@ -58,13 +58,18 @@ def failed
...
+
     return if @participant.nil? || @notifiable.previous_row_num == @notifiable.row_num
 
     # get similar unread notification of challenge
     existing_notification = @participant.notifications.where(notification_type: 'leaderboard', challenge_id: @notifiable.challenge.id, is_new: true)
     existing_notification.delete_all # delete old unread notification of this challenge
 
     message = "You have moved from #{@notifiable.previous_row_num} to #{@notifiable.row_num} place in the #{@notifiable.challenge.challenge} leaderboard"
+
+    return if @participant.notifications.where(notification_type: 'leaderboard', challenge_id: @notifiable.challenge.id, message: message).present?
```
piotrekpasciak, line 71: You need to load all the object to memory here with `where`. Please use `exists?` or `find_by`.

PR #1595, opened 2020-06-23
Added badges display to User Profile

nickinack:


nikhilrayaprolu:
for a quick review for others
image

PR #1600, opened 2020-06-24
Fix notification bell icon issue

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/619640173


app/views/participants/read_notification.js.erb
```diff
@@ -0,0 +1,4 @@
+$('.app-header-notifications').replaceWith("<%= j render partial: 'layouts/notifications_dropdown', locals: { :@notifications => @notifications } %>");
```
piotrekpasciak, line 1: Are you sure this is valid?
`locals: { :@notifications => @notifications }`

Also, please use new hash syntax.

PR #1601, opened 2020-06-24
Versioning feature added for badges

nickinack:
Added Papertrail Versioning for AIcrowd badges!

### DIFF
--- a/app/admin/aicrowd_badges.rb
+++ b/app/admin/aicrowd_badges.rb
@@ -1,3 +1,12 @@
 ActiveAdmin.register AicrowdBadge do
   permit_params :id, :name, :description, :badge_type_id, :code, :badges_event_id
+
+  sidebar "Badge versions", only: [:show, :edit] do
+    ul do
+      aicrowd_badge.versions.reverse.each do |version|
+        li link_to "#{version.created_at}", admin_paper_trail_version_path(version)
+      end
+    end
+  end
+
 end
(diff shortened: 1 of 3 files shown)


app/admin/paper_trail/version.rb
```diff
@@ -3,7 +3,9 @@
 
   show do
     columns do
-      link_to 'Back to Submission', admin_challenge_submission_path(paper_trail_version.item.challenge, paper_trail_version.item)
+      if paper_trail_version.item_type == "Submission"
```
piotrekpasciak, line 6: You can implement the `Back to` link like that:
`link_to "Back to #{paper_trail_version.item_type}", request.referer`

app/admin/aicrowd_badges.rb
```diff
@@ -1,3 +1,23 @@
...
+
+  sidebar "Badge versions", only: [:show, :edit] do
+    ul do
+      aicrowd_badge.versions.reverse.each do |version|
+        li link_to "#{version.created_at}", admin_paper_trail_version_path(version)
+      end
+    end
+  end
+
+  controller do
```
piotrekpasciak, line 12: Are you sure you need this block?  It seems to not be needed after you've used my implementation of PaperTrail versions.

app/models/aicrowd_user_badge.rb
```diff
@@ -1,4 +1,5 @@
 class AicrowdUserBadge < ApplicationRecord
+  has_paper_trail
```
piotrekpasciak, line 2: You didn't implement admin page for these.  Is it needed?

nikhilrayaprolu:
Hey can you add some screenshots!

PR #1602, opened 2020-06-24
Use organizer username as Discourse API username when creating topic

piotrekpasciak:
Monday: 
https://aicrowd.monday.com/boards/363093650/pulses/574264254

Screenshot 2020-06-25 at 00 16 05

### DIFF
--- a/app/services/discourse/create_category_service.rb
+++ b/app/services/discourse/create_category_service.rb
@@ -11,17 +11,27 @@ module Discourse
 
       with_discourse_errors_handling do
         response = create_category_request(ensure_uniqueness: retry_count.positive?)
-
         challenge.update!(discourse_category_id: response.body['category']['id'])
 
         success
       rescue Discourse::UnprocessableEntity => e
-        raise e if retry_count.positive?
+        raise e if retry_count >= 2
 
         retry_count += 1
 
         retry if e.message.include?('Category Name has already been taken')
         retry if e.message.include?('Slug is already in use')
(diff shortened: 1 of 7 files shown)


app/services/discourse/create_category_service.rb
```diff
@@ -1,8 +1,9 @@
 module Discourse
   class CreateCategoryService < ::Discourse::BaseService
     def initialize(challenge:)
-      @client    = prepare_http_client
-      @challenge = challenge
+      api_username = challenge.organizers.first&.participants&.first&.name
```
skbly7, line 4: I agree with the approach.
But it may cause Rollbar error when the organizers has not logged in to Discourse ever.

You should add a fallback to `aicrowd-bot` if possible, OR we can take up AIcrowd <-> Discourse sync immediately. i.e. user creation and username change, etc.

piotrekpasciak: Fallback is implemented here:
https://github.com/AIcrowd/AIcrowd/pull/1602/files#diff-b90063c910e992ed4d4535612c1de784R38

Sync of all participants sounds also as a good. idea.

skbly7: No, this is fallback if the organiser isn't present.

I am saying, when organiser is present BUT his/her account isn't created on Discourse. So on API call you will get some error like user not found, etc.

piotrekpasciak: Sure, will implement retry functionality.

PR #1605, opened 2020-06-25
Fix baseline condition

alkeshrails:
for top three leaderbaords

### DIFF
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -17,7 +17,7 @@ class LeaderboardsController < ApplicationController
         @submitter_submissions.merge!("submitter#{submitter.id}_submissions_by_day": submitter_submissions(submitter).group_by_created_at) if submitter.present?
       end
     end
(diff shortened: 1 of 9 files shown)


app/controllers/leaderboards_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
         @submitter_submissions.merge!("submitter#{submitter.id}_submissions_by_day": submitter_submissions(submitter).group_by_created_at) if submitter.present?
       end
     end
-    @top_three_winners = @leaderboards.where(baseline: nil).first(3)
+    @top_three_winners = @leaderboards.where(baseline: false).first(3)
```
skbly7, line 20: Please run the command on production, to make `baseline: false` wherever it is `baseline: nil` OR better add default `baseline = false` and mark it as compulsory field.

Current your commit would not work on, for example, Blitz, and any older challenges, etc, where old leaderboard computation job has been used for computation (it used to put `baseline: nil`).

PR #1608, opened 2020-06-30
added a task to take weights from csv and add it to challenges

TheCharmingSociopath:


lib/tasks/challenges.rake
```diff
@@ -1,3 +1,5 @@
+require 'csv'
```
skbly7, line 1: You don't need to `require` ideally I guess.

TheCharmingSociopath: Updated

PR #1609, opened 2020-07-01
Create ML challenge and set day of problem

alkeshrails:
Create ML challenge with days limit : N days of ML

https://docs.google.com/document/d/1krhbByFZxyUeqfC9IZsK4w2BA28DXR4mewmqSrSjyDc/edit#heading=h.vy9pu42nawu8

Run rack to create record : `rake create_daily_practice_goal:seed_record`
run to create ActivityPoint record `rake activity_point:create_seed_activity`

PR #1612, opened 2020-07-03
Feature/n days of ml challenge 1 leaderboard calculation

alkeshrails:


alkeshrails:
Will create new table and manage in PR #1609

PR #1614, opened 2020-07-03
Fix foreign key violation issue for participants

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/634019513

rake task to destroy participant associative records if participant does not exists anymore:
rake participant_associative:destroy


app/models/participant.rb
```diff
@@ -31,6 +31,7 @@ class Participant < ApplicationRecord
   after_create :set_email_preferences
   after_save :publish_to_prometheus
   after_commit :upsert_discourse_user, on: [:create, :update]
+  before_destroy :destroy_team_members
```
piotrekpasciak, line 34: It would be better to define relation for `team_members` and use `dependent: :destroy` option.

PR #1632, opened 2020-07-13
Issue : Leaderboard not showing for login user

alkeshrails:
https://aicrowd.monday.com/boards/363093650/pulses/644084787
Show outgoing record to organizer in case of freeze_round

### DIFF
--- a/app/controllers/leaderboards_controller.rb
+++ b/app/controllers/leaderboards_controller.rb
@@ -103,7 +103,7 @@ class LeaderboardsController < ApplicationController
       DisentanglementLeaderboard
         .where(challenge_round_id: @current_round)
         .freeze_record(current_participant)
-    elsif post_challenge? || policy(@challenge).edit? || current_participant&.admin
+    elsif post_challenge? || freeze_record_for_organizer
       policy_scope(OngoingLeaderboard)
         .where(filter)
     else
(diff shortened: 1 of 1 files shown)


app/controllers/leaderboards_controller.rb
```diff
@@ -103,7 +103,7 @@ def set_leaderboards
       DisentanglementLeaderboard
         .where(challenge_round_id: @current_round)
         .freeze_record(current_participant)
-    elsif post_challenge? || policy(@challenge).edit? || current_participant&.admin
+    elsif post_challenge? || freeze_record_for_organizer
```
skbly7, line 106: Can `freeze_record` policy be used instead of this purpose, so the logic of freeze remains at one place?

app/controllers/leaderboards_controller.rb
```diff
@@ -132,4 +132,10 @@ def submitter_submissions(submitter)
   def is_disentanglement_leaderboard?(leaderboard)
     leaderboard.class.name == 'DisentanglementLeaderboard'
   end
+
+  def freeze_record_for_organizer
+    return false unless @current_round&.freeze_flag
```
skbly7, line 137:
```
return false unless challenge_round.freeze_flag && freeze_time(challenge_round)
```

where `freeze_time` is something like:

```
    def freeze_time(ch_round)
      return false if ch_round.end_dttm.nil? || (ch_round.end_dttm - Time.now.utc).negative?

      ch_round.end_dttm - Time.now.utc < ch_round.freeze_duration * 60 * 60
    end
```

skbly7: We calculate `OngoingLeaderboard` always, so it is fine either way i.e. using freeze time or not, you can ignore this then.

PR #1648, opened 2020-07-21
Fix challenge statistics in challenge masthead

piotrekpasciak:
Monday: 
https://aicrowd.monday.com/boards/363093650/pulses/649651556

Screenshot 2020-07-21 at 15 03 33

### DIFF
--- a/app/helpers/challenges_helper.rb
+++ b/app/helpers/challenges_helper.rb
@@ -52,18 +52,18 @@ module ChallengesHelper
     "Draft" if challenge.status == :draft || challenge.status == :starting_soon
   end
 
-  def challenge_stat_count(challenge, stat_type)
+  def challenge_stat_count(challenge, stat_type, parent_meta_challenge = nil)
     case stat_type
     when 'submission'
-      formatted_count(challenge.submissions.count)
+      challenge_submissions_count(challenge, parent_meta_challenge)
     when 'view'
-      formatted_count(challenge.page_views)
+      challenge_page_views_count(challenge, parent_meta_challenge)
     when 'participant'
-      formatted_count(challenge.challenge_participants.size)
+      challenge_participants_count(challenge, parent_meta_challenge)
     when 'vote'
-      formatted_count(challenge.vote_count)
+      challenge_votes_count(challenge, parent_meta_challenge)
     when 'team'
-      formatted_count(challenge.teams_count)
+      challenge_teams_count(challenge, parent_meta_challenge)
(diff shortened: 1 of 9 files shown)


app/models/challenge.rb
```diff
@@ -40,6 +40,8 @@ class Challenge < ApplicationRecord
   has_many :ongoing_leaderboards, class_name: 'OngoingLeaderboard'
 
   has_many :challenge_problems, foreign_key: "challenge_id", class_name: "ChallengeProblems"
+  has_one :parent_challenge_problem, foreign_key: "problem_id", class_name: "ChallengeProblems"
+  has_one :parent_meta_challenge, through: :parent_challenge_problem, source: :challenge, class_name: 'Challenge'
```
skbly7, line 44: We need to use `meta_challenge_id` param (used in set_challenge) instead of relation, because one problem can be part of multiple meta_challenges, and logic changes accordingly

piotrekpasciak: @skbly7 I've handled described scenario, can you Code Review again?

app/helpers/challenges_helper.rb
```diff
@@ -336,12 +336,72 @@ def challenge_dataset_files_path(*args)
...
+    return (challenge_remaining_text(challenge, challenge_round) == "Completed") ? "" : challenge_round.end_dttm
+  end
+
+  private
+
+  def challenge_submissions_count(challenge, parent_meta_challenge)
+    if challenge.meta_challenge?
+      all_challenge_ids = [challenge.id] + challenge.challenge_problems.pluck(:problem_id)
+
+      formatted_count(Submission.where(challenge_id: all_challenge_ids).count)
```
skbly7, line 354: Add filter `meta_challenge_id: challenge.id`

skbly7: And remove `[challenge.id]` from all_challenge_ids

app/helpers/challenges_helper.rb
```diff
@@ -336,12 +336,72 @@ def challenge_dataset_files_path(*args)
...
+    elsif parent_meta_challenge.present?
+      all_challenge_ids = [parent_meta_challenge.id] + parent_meta_challenge.challenge_problems.pluck(:problem_id)
```
skbly7, line 356: In this case:
```
formatted_count(Submission.where(challenge_id: challenge.id, meta_challenge_id: parent_meta_challenge.id).count)
```

PR #1671, opened 2020-08-04
Export Ongoing Leaderboard

alkeshrails:
Export Freeze record in the leaderboard CSV


app/controllers/leaderboards_controller.rb
```diff
@@ -44,7 +44,7 @@ def index
   def export
     authorize @challenge, :export?
 
-    @leaderboards = Leaderboard
+    @leaderboards = OngoingLeaderboard
```
skbly7, line 47: Condition for freeze_leaderboard check?

PR #1713, opened 2020-10-20
fixes leaderboard base line

shell-x:
image (1)


skbly7:
Thanks for finding the fix. 
We unknowingly fixed it when creating new field in this PR: #1711, due to which closing this PR.

PR #1715, opened 2020-10-23
pin user rank on challenges leaderboard

### LINKED ISSUE #1691: Pin/stick the user's rank on the challenge's leaderboard
It will be good to have a sticky row showing the participant's rank on the leaderboard (basically should be visible always). This is already happening on the site wide leaderbord. It will be good to have it on the leaderboards of individual challenges as well.

shell-x:
fixes #1691


shell-x:
working in UAT environment

ezgif com-optimize

PR #1726, opened 2020-11-07
Add feature- export participants csv

shell-x:


app/services/challenge_participants/csv_export_service.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      end
+      success(csv_data)
+    end
+
+    private
+
+    def csv_headers
+      headers = [
+          'Participant Id',
```
skbly7, line 32: ID doesn't need to be exported

app/services/challenge_participants/csv_export_service.rb
```diff
@@ -0,0 +1,77 @@
...
+      @challenge_participants = challenge_participants
+      @challenge = challenge
+    end
+
+    def call
+      csv_data = CSV.generate(headers: true) do |csv_file|
+
+        csv_file << csv_headers
+
+        @challenge_participants.find_each do |participant|
```
skbly7, line 13: should use:

```
        @challenge_participants.find_each do |challenge_participant|
        participant = challenge_participant.participant
        [....]
```

app/services/challenge_participants/csv_export_service.rb
```diff
@@ -0,0 +1,77 @@
...
+          teams = get_participant_teams(@challenge, participant.participant_id)
+          if teams.present? then
+            teams.each do |_team|
```
skbly7, line 16: One participant would never be part of multiple teams. You can just do single team thing.

skbly7 (changes requested):

PR #1727, opened 2020-11-07
Add searchable dropdown to admin/blogs

shell-x:
Add search feature to admin participants dropdown list.
Using gem `activeadmin-searchable_select`. Use attribute `searchable_select` as in admin forms to enable.

participant_selectable_dropdown


app/admin/blog.rb
```diff
@@ -24,7 +24,7 @@ def permitted_params
 
   form do |f|
     f.inputs do
-      f.input :participant
+      f.input(:participant, as: :searchable_select)
```
skbly7, line 27: Need to be done at all admin forms

skbly7 (changes requested):
^

PR #1730, opened 2020-11-17
create unique name for images

sujnesh:


app/uploaders/raw_image_uploader.rb
```diff
@@ -10,4 +10,12 @@ def store_dir
   def extension_white_list
     ['jpg', 'jpeg', 'gif', 'png', 'svg']
   end
+
+  def filename
+    random_token = Digest::SHA2.hexdigest("#{Time.now.utc}--#{model.id.to_s}").first(20)
+    ivar = "@#{mounted_as}_secure_token"
+    token = model.instance_variable_get(ivar)
+    token ||= model.instance_variable_set(ivar, random_token)
+    "#{model.id}_#{token}.jpg" if original_filename
```
skbly7, line 19: Maybe use `#{file.extension}` instead of hard coding to jpg here?
Reference: https://github.com/carrierwaveuploader/carrierwave/wiki/how-to:-create-random-and-unique-filenames-for-all-versioned-files

skbly7: You can also remove `model.id`, it can potentially make internal IDs for challenge, participants, etc public, which doesn't seem necessary.

skbly7 (changes requested):

skbly7 (approved):

PR #1738, opened 2020-11-30
added default page value and changed arg values

sujnesh:


skbly7 (approved):
Made minor change and approving

PR #1748, opened 2020-12-06
remove username reset for github login and fix non unique username bug

sujnesh:


app/controllers/participants/omniauth_callbacks_controller.rb
```diff
@@ -49,13 +52,17 @@ def from_omniauth(auth)
...
+    username = userhandle.to_ascii
       .tr("@", "a")
       .gsub("&", "and")
       .delete('#')
       .delete('*')
       .gsub(/[\,\.\'\;\-\=]/, "")
       .gsub(/[\(\)]/, "_")
       .tr(' ', "_")
+
+    while Participant.where(name: username).exists?
```
skbly7, line 64: return nil otherwise or some error, instead of username + 9?

skbly7 (changes requested):

skbly7 (approved):

app/controllers/participants/omniauth_callbacks_controller.rb
```diff
@@ -29,6 +29,9 @@ def failure
 
   def from_omniauth(auth)
     email     = auth.info.email
+    participant = User.find_by(email: email)
```
skbly7, line 32: `Participant.find_by`

skbly7 (changes requested):

skbly7 (approved):

PR #1752, opened 2020-12-08
added default leaderboard for column migrated from challenge round to…

sujnesh:
… leaderboard


app/models/concerns/freeze_record.rb
```diff
@@ -7,7 +7,7 @@ def self.freeze_record(participant)
 
       ch_round = first.challenge_round
       if freeze_duration?(participant)
-        freeze_beyond_time = ch_round.end_dttm - ch_round.freeze_duration.to_i.hours
+        freeze_beyond_time = ch_round.end_dttm - ch_round.default_leaderboard.freeze_duration.to_i.hours
```
skbly7, line 10: Check where is it being used, if that place has `challenge_leaderboard_extra` or `leaderbooard` param, you might want to use that instead of `challenge_round&.default_leaderboard`

skbly7 (changes requested):

skbly7 (approved):

PR #1793, opened 2021-01-16
add explicit return statement

sujnesh:


skbly7 (approved):
Ruby 🤷 

For future not to self/us:

```
(production) > def a
  1
  2 if 1==2
end

(production) > a()
=> nil
```

PR #1795, opened 2021-01-18
leaderboard sequencing/choose default leaderboard

sujnesh:


app/controllers/challenges_controller.rb
```diff
@@ -323,6 +324,19 @@ def get_date_wise_challenge_problem
     @challenge.problems.where(id: problem_ids)
   end
 
+  def update_default_leaderboard
```
skbly7, line 327: This all may not be required...

app/views/insights/_charts/_top_score_vs_time.html.erb
```diff
@@ -15,7 +15,7 @@
       <% end %>
     </div>
       <% challenge_rounds.each do |challenge_round| %>
-        <% if challenge_round.default_leaderboard.score_secondary_title.present? %>
+        <% if challenge_round.default_leaderboard.score_secondary_title.present? && challenge_round.default_leaderboard.secondary_sort_order_cd != 'not_used' %>
```
skbly7, line 18: Let's use `hidden` instead of `not_used`, my bad in sharing specs here.

app/views/challenges/fields/_challenge_round_fields.html.erb
```diff
@@ -38,6 +38,11 @@
           <%= f.label :end_dttm, 'Round End (UTC)' %>
           <%= f.text_field :end_dttm, class: "form-control", data: {controller: 'flatpickr'} %>
         </div>
+
+        <div class="form-group col-sm-12 col-md-6">
```
skbly7, line 42: It should be part of leaderboard not challenge round settings.

app/views/challenges/fields/_leaderboard_fields.html.erb
```diff
@@ -13,6 +13,9 @@
       <% if f.object.default %>
         <span class="float-left"><span class="fa fa-star-o fa-3x"></span> Default</span>
       <% end %>
+      <%= render partial: 'challenges/fields/fancy_radio', locals: { f: f,
```
skbly7, line 16: We can have similar jQuery for selection of default one. So toogle on this automatically make other leaderboards in this Round as "not" default.
https://github.com/AIcrowd/AIcrowd/blob/master/app/javascript/stimulus_js/controllers/challenges_edit_controller.js#L16

skbly7: And you can then directly put `default` as variable instead of `is_default_leaderboard` and the code added in the controller for default thing.

skbly7 (changes requested):

skbly7 (approved):

skbly7 (approved):

PR #1818, opened 2021-02-04
gitlab user creation api integration

sujnesh:


app/services/gitlab/fetch_user_by_external_id_service.rb
```diff
@@ -22,7 +39,11 @@ def call
     attr_reader :client, :participant
 
     def endpoint_path
-      "api/v4/users?extern_uid=#{participant.id}&provider=oauth2_generic"
+      "api/v4/users?username=#{participant.name}"
```
skbly7, line 42: no, it should be `extern_uid` only, it is more reliable and pretty sure way to know the mapping

app/services/gitlab/fetch_user_by_external_id_service.rb
```diff
@@ -10,6 +10,23 @@ def call
...
+        if user_data.blank?
+          response = client.post(
+            new_user_endpoint_path,
+            {
+              email: participant.email,
+              name: participant.first_name + ' ' + participant.last_name,
+              username: participant.name,
+              skip_confirmation: true,
+              force_random_password: true,
+              reset_password: true
```
skbly7, line 22: Not needed, we don't want users to get into a login/password mess (reset -> will send email), and they should just be able to use oAuth login with AIcrowd.

It basically needs to remain transparent.

skbly7: Ah, I see what's happening, let's start sending `extern_uid: participant.id` here.

skbly7 (changes requested):

PR #1825, opened 2021-02-11
home page caching

sujnesh:


app/controllers/application_controller.rb
```diff
@@ -30,7 +29,9 @@ def append_info_to_payload(payload)
   end
 
   def notifications
-    @notifications = current_user&.notifications
+    @notifications = Rails.cache.fetch('participant_notifications', expires_in: 5.minutes) do
```
sujnesh, line 32: Per user

app/views/shared/challenges/_listing.html.erb
```diff
@@ -1,110 +1,112 @@
...
-      <%= link_to challenge_path(challenge) do %>
-        <div class="card card-challenge">
-      <% end %>
-    <% else %>
-      <div class="card card-challenge" style="background-image: url('<%= challenge.image_url %>');background-size: 100% auto;">
-    <% end %>
-      <div class="image-wrapper" style="padding-bottom: 56.2%;position: relative;">
-        <% if !challenge.big_challenge_card_image? %>
-          <%= image_tag(challenge.image_url, class: 'card-img image_16x9') %>
+  <% cache (challenge) do %>
```
sujnesh, line 2: How does view count work here. Either add timeout or find better invalidation.

db/migrate/20210210191019_create_challenge_properties.rb
```diff
@@ -0,0 +1,10 @@
+class CreateChallengeProperties < ActiveRecord::Migration[5.2]
+  def change
+    create_table :challenge_properties do |t|
+      t.integer :page_views
```
sujnesh, line 4: default and required true here

PR #1852, opened 2021-03-05
File size helper changed for nil

skbly7:


app/helpers/files_helper.rb
```diff
@@ -17,14 +17,16 @@ def file_title(file)
 
   def file_size(file)
     if file.try(:hosting_location) == 'Own S3' || file.new_record?
-      number_to_human_size(file.external_file_size.to_i)
+      return number_to_human_size(file.external_file_size.to_i)
     elsif file.try(:hosting_location) == 'External'
-      file.external_file_size
+      if file.external_file_size.present?
```
sujnesh, line 22: This will return nil in case the condition fails. Is that okay?

sujnesh (approved):

PR #1868, opened 2021-03-19
added challenge leaderboard in meta challenge

sujnesh:


app/services/calculate_meta_leaderboard_service.rb
```diff
@@ -61,7 +65,8 @@ def participant_values(rank, participant, scores)
       seq: rank,
       meta: scores['rank'],
       submitter_type: participant[0],
-      submitter_id: participant[1]
+      submitter_id: participant[1],
+      challenge_leaderboard_extra_id: @challenge_leaderboard_extra.id
     }
   end
```
skbly7, line 72:
```
    BaseLeaderboard.where(challenge_id: @challenge.id, challenge_round_id: @round.id).delete_all
```

fix it to include challenge_leaderboard_extra_id

app/services/calculate_meta_leaderboard_service.rb
```diff
@@ -1,12 +1,16 @@
 class CalculateMetaLeaderboardService
 
-  def initialize(challenge_id:)
+  def initialize(challenge_id:, challenge_leaderboard_extra: nil)
```
skbly7, line 3: change calculateLeaderboardJob too, so it can start calling for all leaderboards instead of only default

skbly7 (changes requested):

app/services/calculate_meta_leaderboard_service.rb
skbly7: take example from normal leaderboard one.

- in case of default leaderboard: delete challenge_leaderboard_extra_id & nil
- in case of other leaderboard: delete challenge_leaderboard_extra_id

app/jobs/calculate_leaderboard_job.rb
```diff
@@ -23,7 +25,11 @@ def perform(challenge_round_id:)
       challenge_round.challenge_leaderboard_extras.each do |challenge_leaderboard_extra|
         ChallengeRounds::CreateLeaderboardsService.new(challenge_round: challenge_round, meta_challenge_id: challenge_problem.challenge_id, challenge_leaderboard_extra: challenge_leaderboard_extra).call
       end
-      CalculateMetaLeaderboardService.new(challenge_id: challenge_problem.challenge_id).call
+
+      challenge_problem.challenge.active_round.challenge_leaderboard_extras.each do |challenge_leaderboard_extra|
+        CalculateMetaLeaderboardService.new(challenge_id: challenge_problem.challenge_id, challenge_leaderboard_extra_id: challenge_leaderboard_extra).call
```
skbly7, line 30: Do you mean `challenge_leaderboard_extra`? ;)

app/services/calculate_meta_leaderboard_service.rb
```diff
@@ -61,13 +65,16 @@ def participant_values(rank, participant, scores)
...
       submitter_type: participant[0],
-      submitter_id: participant[1]
+      submitter_id: participant[1],
+      challenge_leaderboard_extra_id: @challenge_leaderboard_extra.id
     }
   end
 
   def create_leaderboard
     people = {}
+    filter = @child_leaderboards&.first&.first&.challenge_leaderboard_extra&.filter
```
skbly7, line 75: what's happening here?

PR #1873, opened 2021-03-31
Image processing in background

sujnesh:


Gemfile
```diff
@@ -108,6 +108,7 @@ gem 'derailed'
 gem 'barnes'
 gem 'gemoji-parser'
 gem 'clipboard-rails'
+gem 'carrierwave_backgrounder', git: 'https://github.com/lardawge/carrierwave_backgrounder'
```
skbly7, line 111: forked it to AIcrowd namespace, let's use it for stability:
https://github.com/AIcrowd/carrierwave_backgrounder

skbly7: looks like they have gem published too, any reason why we are using `git` at first place?

PR #1886, opened 2021-04-12
Tech 376

sujnesh:


app/controllers/submissions_controller.rb
```diff
@@ -106,6 +106,10 @@ def show
   end
 
   def new
+    if @challenge.min_team_participants > 1 && (@challenge.min_team_participants > get_team_participants.count)
+      redirect_to @challenge,
```
skbly7, line 110: `redirect_or_json` need to be used

app/controllers/concerns/challenge_masthead.rb
```diff
@@ -8,6 +8,7 @@ def challenge_masthead_actions(actions)
         before_action :set_challenge_rounds, only: actions
         before_action :set_vote, only: actions
         before_action :set_follow, only: actions
+        before_action :redirect_if_hidden, only: actions
```
skbly7, line 11: Not sure why it is here? Should it be part of Discussions controller?

skbly7 (changes requested):

app/policies/challenge_policy.rb
```diff
@@ -159,6 +160,13 @@ def clef_participant_registered?(challenge)
     end
   end
 
+  def get_team_participants_count
+    team = participant.teams.where(challenge_id: @record.id).first
```
skbly7, line 164: what if there is no team?

sujnesh, line 164: Will break if no team

skbly7:
```
if team.present?
    participants = team.team_participants.map(&:participant) if team.present?
    participants = [participant] if participants.blank?
    return participants.count
end
return 1
```

skbly7 (approved):

PR #1897, opened 2021-04-26
removed jquery and popper link and added basline api

sujnesh:


app/models/challenge.rb
```diff
@@ -71,6 +71,7 @@ class Challenge < ApplicationRecord
   has_many :posts
   has_many :locked_submissions
   has_one :challenge_property
+  has_one :baseline
```
skbly7, line 74: `has_many`

skbly7: `default` column

skbly7 (approved):

skbly7 (approved):

PR #1945, opened 2021-06-16
added captcha to email related forms

sujnesh:


app/controllers/participants/passwords_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Participants::PasswordsController < Devise::PasswordsController
+  before_action :check_captcha, only: [:create]
+
+  def check_captcha
+    byebug
```
skbly7, line 5: We need to say bye to this bug.... 🤷

app/controllers/challenge_call_responses_controller.rb
```diff
@@ -51,4 +52,10 @@ def challenge_call_response_params
   def set_challenge_call
     @challenge_call = ChallengeCall.friendly.find(params[:challenge_call_id])
   end
+
+  def check_captcha
+    unless verify_recaptcha
+      redirect_to root_path
```
skbly7, line 58: Can we redirect with some error message, and back to same page (with form fields remaining filled ideally)? 
This is an important form and entry point for new challenges.

```
We are unable to validate this request, in case the error continues please email us at hello@aicrowd.com.
```

skbly7 (changes requested):

skbly7 (approved):

PR #1946, opened 2021-06-18
Revamped notebook design

sujnesh:


app/controllers/post_bookmarks_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  def create
+    PostBookmark.create!(post_id: params[:post_id], participant_id: current_participant.id)
+
+    respond_to do |format|
+      format.js { render :refresh }
+    end
+  end
+
+  def destroy
+    PostBookmark.where(id: params[:id]).first.destroy
```
skbly7, line 15: Authorisation missing, `PostBookmark.participant_id == current_participant.id`, add in policy or here.

app/controllers/posts_controller.rb
```diff
@@ -172,5 +178,10 @@ def update_post_categories
         end
       end
     end
+
+    def set_page_view
```
skbly7, line 182: NO!!!!!!

Add field in the `Post` and populate data once with ahoy_query, BUT DO NOT RUN IT IN PRODUCTION, THINGS WILL CRASH

app/models/post.rb
```diff
@@ -27,6 +28,10 @@ def vote(participant)
     self.votes.where(participant_id: participant.id).first if participant.present?
   end
 
+  def bookmark(participant)
+    self.post_bookmarks.where(participant_id: participant.id).first if participant.present?
```
skbly7, line 32: `self.post_bookmarks.where(participant_id: participant.id).count` will be faster and cleaner

skbly7:
I have not checked the frontend part, the backend has a lot of fixes to do.

PR #1957, opened 2021-07-12
Mixpanel integration

chittalpatel:
Mixpanel setup, Mixpanel ruby jobs implementation, Mixpanel browser initial work by @skbly7 
Fix mixpanel jobs, fix login/registration and participation flow by @chittalpatel 

- Track anonymous users (not signed in)
- Track page visits
- Track login/registration flow
- Track challenge participation flow

### DIFF
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -32,7 +32,6 @@ class ParticipantsController < ApplicationController
     end
     @achievements_count = 0
     @participant.badges.badges_stat_count.map { |badge_type_id, badge_type_count| @achievements_count += badge_type_count if [1,2,3].include?(badge_type_id) }
-    Mixpanel::SyncJob.perform_later(@participant)
   end
 
   def edit; end
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -32,6 +32,7 @@ class ParticipantsController < ApplicationController
     end
     @achievements_count = 0
     @participant.badges.badges_stat_count.map { |badge_type_id, badge_type_count| @achievements_count += badge_type_count if [1,2,3].include?(badge_type_id) }
+    Mixpanel::SyncJob.perform_later(@participant)
   end
 
   def edit; end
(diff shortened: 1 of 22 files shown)


app/controllers/participants_controller.rb
```diff
@@ -32,6 +32,7 @@ def show
     end
     @achievements_count = 0
     @participant.badges.badges_stat_count.map { |badge_type_id, badge_type_count| @achievements_count += badge_type_count if [1,2,3].include?(badge_type_id) }
+    Mixpanel::SyncJob.perform_later(@participant)
```
skbly7, line 35: Reason to add here?

app/javascript/stimulus_js/controllers/mixpanel_controller.js
```diff
@@ -0,0 +1,78 @@
...
+        mixpanel.track('Logout', {
+            "Logged In": false
+        }, function () {
+            mixpanel.reset();
+        })
+    }
+
+    challenge_participation_start() {
+        mixpanel.track('Challenge Participation Start');
+    }
```
skbly7, line 77: `Challenge Participation Complete` event is missing?

app/controllers/participants_controller.rb
chittalpatel: We will be tracking unique users but also we need to create user profiles for these users... Location of doing this is debatable so I've kept here because user will go to his page profile page while using AIcrowd then his profile will be set... 
To create profiles quickly, we could also add this in controller which handles login which would we be more frequent event then going to user profile page...

app/javascript/stimulus_js/controllers/mixpanel_controller.js
chittalpatel: It is implemented in `app/controllers/challenge_participants_controller.rb: line 34`

app/controllers/participants_controller.rb
skbly7: Let's do it in `models/participant.rb` as `after_save` hook.
Example:
https://github.com/AIcrowd/AIcrowd/blob/master/app/models/participant.rb#L32

This place isn't good, because it will be called N times i.e. anyone visits the profile page and cause unnecessary SyncJob.

skbly7 (approved):

PR #2030, opened 2021-12-20
Weekly challenge popup

sujnesh:


app/controllers/participants_controller.rb
```diff
@@ -147,6 +147,15 @@ def switch_tab
     end
   end
 
+  def interacted_with_popup
```
skbly7, line 150: Should be a javascript thing!

PR #2038, opened 2022-01-12
Search refactor

sujnesh:


sujnesh:
To add a new model after deployment; in console run -->
PgSearch::Multisearch.rebuild(Participant) Here Participant is to be replaced by the model name we want.
For more details: https://github.com/Casecommons/pg_search#rebuilding-search-documents-for-a-given-class

PR #2093, opened 2022-04-11
Added images for the auth page of AIcrowd API

gauranshkumar:
Added SVG for using in design template of /auth page and other pages in AIcrowd API and Datasets


gauranshkumar:
@skbly7 can you review it!

